PR #10, opened 2020-09-10
Update package metadata

whymarrh:
This change updates all of the package metadata.

- [x] Use the correct license identifier
- [x] Move all dev dependencies into `devDependencies`
- [x] Add a `publishConfig` and add a package scope (`@metamask`)
- [x] Add a explicit `files` list


Gudahtt (approved):
LGTM!

PR #17, opened 2020-09-17
Add LavaMoat

### LINKED ISSUE #15: Add Lava-Moat
How cool would it be if every MetaMask module shipped with its dependencies pre-confined by default via [lava moat](https://github.com/LavaMoat/LavaMoat#bundler-plugins)?

Problem currently this template repo uses rollup as its build system, but Lava Moat only has plugins for browserify and webpack. This means we either need to switch this project to one of those ([like this PR that converts it to webpack](https://github.com/MetaMask/metamask-module-template/pull/16)), or add a lava-moat plugin for rollup.

[one guide on creating a rollup plugin](https://lihautan.com/12-line-rollup-plugin/)

danfinlay:
Fixes #15 (Add lavamoat) by integrating #16 to use webpack.

Inspecting the bundle this seems to have worked quite easily!


danfinlay:
I kinda think "add lava moat to a module template" is an excellent way to use it. The hardest part about lava moat is probably applying it to a legacy project that might have dependencies that are for unknown reason incompatible. This encourages selecting dependencies in part because of their lavamoat compatibility.

kumavis:
this is neat!
as we discussed, should focus.on running the node process under lavamoat-node first
the webpack stuff in general is not recommended for use

Issue #3, opened 2020-09-04
Add npm publishing to the CI post-merge hook


danfinlay:
Depends on #2

rekmarks:
Closing in favor of #31

PR #19, opened 2020-09-23
Update GitHub CI workflow

whymarrh:
This change updates the GitHub Actions workflow used for CI, adding names to the jobs as well as a terminal job that will simplify the process of adding required status checks to branches.

Having a known terminal job that is consistent across all projects make it easy to add a required status check for a branch—one that will not change as new intermediate jobs are added.

I've added a lint step and the flags from our existing CircleCI config as well.


whymarrh:
I've created #20 to represent the work required to get the lint command going again

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,32 @@
+name: Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
Gudahtt, line 7: Is there any harm in removing this, and running on the `pull_request` event for any branch? CI is useful sometimes for making changes to feature branches

```suggestion
```

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,32 @@
...
+    - run: yarn build
+    - run: yarn lint
+    - run: yarn test
+  all-jobs-pass:
+    name: All jobs pass
+    runs-on: ubuntu-20.04
+    needs:
+      - build
+    steps:
+    - uses: actions/checkout@v2
```
Gudahtt, line 32: Could this step me omitted? :thinking: I don't see much utility in checking out the repo here.

whymarrh: There needs to be _some_ step and this seems like a harmless empty step

Gudahtt: OK, fair enough. I wasn't aware a step was required. We can replace this later with something more lightweight if need be.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,31 @@
+name: Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
```
Gudahtt, line 6: Note that this does need to be `pull_request_target` if any of our action workflows require secrets, since those won't be provided on forks :thinking: 

`pull_request` seems perfectly fine to use on repos _without_ secrets though - maybe even preferable, as testing changes to the workflows is easier with `pull_request`. So maybe this is fine as a default.

Gudahtt (approved):
LGTM!

.github/workflows/build-test.yml
danfinlay: Since [we want to add npm publishing](https://github.com/MetaMask/metamask-module-template/issues/3) eventually, this seems like an important suggestion, though I'm not sure it needs to block this PR.

Gudahtt: Publishing would run post-merge, so that would be from the `push` event. I don't think using `pull_request` poses any challenges there.

My comment was more about status checks that use secrets, such as on `metamask_extension` where our CI builds have valid non-production Sentry and Matomo keys.

PR #29, opened 2021-03-23
Update per de facto standard

rekmarks:
This PR rewrites the module template as though it were rewritten to use our de facto standard, with the following arguable exceptions:
- Use of GitHub Actions remains (we currently use CircleCI in practice)
- Use of `ts-jest` remains (we haven't started using that in most places, because we usually haven't bothered migrating tests to TypeScript)

Consider this an RFC of sorts. I think it would be helpful to restart discussions around this repo by starting with what we have and working from there.

Relative to `main`, this PR:
- Removes `rollup` in favor of plain `tsc`
  -  We don't use `rollup` in the way we did in this module repo anywhere else. 
- Moves build output to `dist` folder
  - When the time comes to actually decide what to do, if I had to pick a hill to die on, this would be it.
- Adds `eslint`, `@metamask/eslint-config`, and relevant dependencies
- Adds MetaMask's proprietary license
  - Per ConsenSys's licensing policy, new repositories are only open source with legal's permission
- Adds an `.nvmrc` with `v12`
  - Node 10 is EOLing, and IIRC bumping to 12 is a tooling week item
- Moves the tests to the `test` folder
- `tsconfig.json`
  - Removes `DOM` from `lib` array
  - Updates `target` and `lib` from `es6` and `es2016` to `ES2017`
- Specifies a minimum Node version of `12` via `engines` in `package.json`


.eslintrc.js
```diff
@@ -9,9 +8,14 @@ module.exports = {
     '@metamask/eslint-config/config/typescript',
```
Gudahtt, line 8: I would suggest putting the typescript rules in an override. This config disables a variety of ESLint rules that we still want to apply to JavaScript modules. Also the TypeScript lint rules can be incorrectly triggered by lint errors in JavaScript modules in some cases (I recently encountered this here: https://github.com/MetaMask/controllers/pull/396

.eslintrc.js
```diff
@@ -9,9 +8,14 @@ module.exports = {
...
   ],
 
+  plugins: [
+    'json',
```
Gudahtt, line 12: This is a bad plugin, but I guess it makes sense to keep using it until we add `prettier` to `@metamask/eslint-config` :disappointed: 
`prettier` completely supplants this. It is faster and better at validating JSON files.

.eslintrc.js
```diff
@@ -20,4 +24,10 @@ module.exports = {
       '@metamask/eslint-config/config/nodejs',
     ],
   }],
-};
\ No newline at end of file
+
+  ignorePatterns: [
+    '!.eslintrc.js',
+    'dist/',
+    'node_modules/',
```
Gudahtt, line 31: This entry is not necessary - [`node_modules` is always ignored](https://eslint.org/docs/user-guide/configuring/ignoring-code), along with all dotfiles. That's why the `!.eslintrc.js` entry is required.

LICENSE
```diff
@@ -0,0 +1,18 @@
+Copyright ConsenSys Software Inc. 2021. All rights reserved.
```
Gudahtt, line 1: A better default for this repository would be to have no license at all. We can choose a license for individual repositories as we create them.

README.md
```diff
@@ -33,13 +29,12 @@ To run tests on file changes, run `yarn test:watch`.
 The project follows the same release process as the other libraries in the MetaMask organization:
 
 1. Create a release branch
-    - For a typical release, this would be based on `master`
-    - To update an older maintained major version, base the release branch on the major version branch (e.g. `1.x`)
+  * For a typical release, this would be based on `master`
```
Gudahtt, line 32: We've been migrating to use `main` as the main development branch
```suggestion
  * For a typical release, this would be based on `main`
```

jest.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
       statements: 100,
     },
   },
-  moduleFileExtensions: ['ts', 'tsx', 'json', 'js', 'jsx', 'node'],
+  moduleFileExtensions: ['ts', 'json', 'js', 'node'],
```
Gudahtt, line 12: Note that this is for the file extensions used not just by the project directly, but by _all dependencies_. [The Jest docs suggest moving the most commonly-used extensions to the left](https://jestjs.io/docs/configuration#modulefileextensions-arraystring), but I'd suggest leaving all of the default entries there.

I discovered this when I went to remove `node` from the array on the `@metamask/controllers` repo, and the tests broke! Apparently a dependency is using the `.node` file extension.

tsconfig.json
```diff
@@ -1,17 +1,18 @@
...
   "compilerOptions": {
-    "moduleResolution": "Node",
     "declaration": true,
+    "esModuleInterop": true,
+    "lib": [
+      "ES2017"
+    ],
     "module": "ES6",
+    "moduleResolution": "Node",
+    "outDir": "dist",
```
Gudahtt, line 10: I'd suggest specifying a `rootDir` of `./src`, to ensure that we don't accidentally include a module outside of dir and mess up the resulting directory structure of the `tsc` build.

tsconfig.json
```diff
@@ -1,17 +1,18 @@
 {
   "compilerOptions": {
-    "moduleResolution": "Node",
     "declaration": true,
+    "esModuleInterop": true,
+    "lib": [
```
Gudahtt, line 5: Thoughts on adding `"inlineSources": true,`? It seems like a good default.

test/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import greeter from './index';
+import greeter from '../src';
```
Gudahtt, line 1: This change seems contrary to our de-facto standard :thinking: In practice it has always been a mix of both patterns., but we've been moving toward co-located tests recently. Now tests are co-located on both products and in the controllers repo.

.eslintrc.js
rekmarks: Ah, I see. Well, hopefully we should be able to get `prettier` into the base config soon.

In fact, I think `prettier` will replace a number of `@typescript-eslint` formatting rules as well. `@typescript-eslint` has a lot of technical debt and some rules that are straight up broken (e.g. indentation for generics).

LICENSE
rekmarks: Yeah, I suppose we only need to add a license for the first release of a package. I think there's a form we can use to submit open-sourcing requests to legal. I'll add a todo for myself to check that out.

test/index.test.ts
rekmarks: Hrm. We definitely still keep tests in a separate folder in the majority of our repos. I support the colocation of tests in e.g. the extension because of how big the file tree is, but I like the clean separation between tests and source code for small packages. I could be brought around, but I have to sleep on it, lol.

Gudahtt: Ah, I had thought we were moving to co-located tests, and the separate test directory pattern was around because we hadn't migrated it yet! Hadn't realized that anyone still preferred it, or was still actively migrating to it.

package.json
```diff
@@ -1,32 +1,36 @@
...
-  "description": "",
+  "version": "0.0.0",
+  "description": "The MetaMask Node module template.",
   "publishConfig": {
     "registry": "https://registry.npmjs.org/",
     "access": "public"
   },
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.es.js",
-  "types": "dist/index.d.ts",
```
Gudahtt, line 11: No more types? :thinking: This is recommended by TypeScripts docs: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

rekmarks: Whoops, fat-fingered that one. Restored!

Gudahtt (approved):
LGTM!

PR #30, opened 2021-03-25
Use ES2020 lib in TypeScript

rekmarks:
Updates the `lib` field of `tsconfig.json` to use `ES2020` while still targeting `ES2017`.

We compile our TypeScript source using `tsc` to the `dist` folder, which produces `ES2017`-compatible code, even if we use `ES2020` features in our source code.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "declaration": true,
     "esModuleInterop": true,
     "inlineSources": true,
-    "lib": ["ES2017"],
+    "lib": ["ES2020"],
     "module": "ES6",
     "moduleResolution": "Node",
     "outDir": "dist",


Gudahtt:
>Regarding the ES version, `tsc` can compile down ES2020 code, and we need at least ES2018 in order to do things like `async () => ...`. I don't know of a reason not to use the latest finalized ES spec of it can be compiled down to our target environment(s) by tsc.

We don't compile out dependencies with `tsc`, nor do we run them through Babel. We'd have to go our of our way to do that, and it'd slow the build down substantially. We really don't want our libraries to require transpilation.

Also `async / await` is part of ES2017. That's why that has been the minimum set throughout most libraries.

Gudahtt (approved):
LGTM!

PR #33, opened 2021-03-29
Colocate test files with source files

### LINKED ISSUE #32: Decide where tests should live
This repository has tests in a top-level `test` directory, but lately we've been migrating tests to be alongside the code under test (as is done in the controllers repo, the extension, and mobile).

Ideally we'd use the same pattern everywhere of having tests alongside the code under test. If we're not going to do that, we should have some guidance on where we'll follow each pattern, and why.

rekmarks:
Removes the `test` directory and colocates the `index.test.ts` with `index.ts`. Updates `tsconfig.json` to ignore test files in all directories.

Closes #32

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,5 +12,6 @@
     "strict": true,
     "target": "ES2017"
   },
-  "include": ["./src/*.ts"]
+  "exclude": ["./src/**/*.test.ts"],
+  "include": ["./src/**/*.ts"]
 }
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,4 +1,4 @@
-import greeter from '../src';
+import greeter from '.';
 
 describe('Test', () => {
   it('greets', () => {


tsconfig.json
```diff
@@ -14,7 +14,10 @@
     "strict": true,
     "target": "ES2017"
   },
+  "exclude": [
+    "**/*.test.ts"
+  ],
   "include": [
-    "./src/*.ts"
+    "./src/**/*.ts"
```
brad-decker, line 21: why specifying src folder here but exclude has **/*.test.js? Also unrelated: should we add 'node_modules' to exclude?

Gudahtt: It's because tests might include test-specific libraries and mocks that are outside of the `src` directory.

The "main" tests for each module can certainly be co-located, but there is a lot of test-related stuff that doesn't have any clear home in the `src` directory. A `test` directory makes sense for that stuff.

rekmarks: Hm, to Brad's point though, a `test` directory wouldn't be included in the first place. I'll push an update.

Also, `node_modules` is always excluded, IIRC.

rekmarks: Exclude reference: https://www.staging-typescript.org/tsconfig#exclude

brad-decker (dismissed):

Gudahtt (dismissed):

Gudahtt:
@rekmarks This good to merge now?

Gudahtt (approved):
LGTM!

PR #34, opened 2021-03-31
Implement standard dependabot

darkwing:
A few notes:

- `label`'s default value is `dependencies`, so I left that off
- `rebase-strategy`'s default value is `auto`, so I left that off
- `live` is not a supported YML setting, unfortunately, so to get "live" PRs opened when we update our own packages, we'll need to bump each dependabot file individually.

Also note that this config should be pretty portable -- one issue we'll need to rectify is the main branch name on different repositories differs.

### DIFF
--- a/.github/dependabot.yml 
+++ b/.github/dependabot.yml 
@@ -10,3 +10,4 @@ updates:
       time: "06:00"
     target-branch: "main"
     versioning-strategy: "increase-if-necessary"
+    open-pull-requests-limit: 10
--- a/.github/dependabot.yml 
+++ b/.github/dependabot.yml 
@@ -1,0 +1,14 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "main"
+    allow:
+      - dependency-name: "@metamask/*"
+        dependency-type: "development"
+    versioning-strategy: "increase-if-necessary"
--- a/.github/dependabot.yml 
+++ b/.github/dependabot.yml 
@@ -7,8 +7,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+      time: "06:00"
     target-branch: "main"
-    allow:
-      - dependency-name: "@metamask/*"
-        dependency-type: "development"
     versioning-strategy: "increase-if-necessary"


.github/dependabot.yml 
```diff
@@ -0,0 +1,14 @@
...
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "main"
+    allow:
+      - dependency-name: "@metamask/*"
+        dependency-type: "development"
```
darkwing, line 13: Comment from @Gudahtt: `I'm curious why this is here. I would think we'd want all updates to our packages, not just those that are devDependencies. 🤔`

darkwing:
Comment from @Gudahtt: 

> This configuration will only update our own packages. Any thoughts on using Dependabot to flag updates to third party dependencies that we're using as well?

Response from @brad-decker:

> I think we should. Sometimes shipping a dependency update yields great improvements for our users. Sometimes it can also fubar the whole thing, but hopefully as we extend E2E test coverage we'll catch those kind of issues in the dependabot prs

.github/dependabot.yml 
```diff
@@ -0,0 +1,14 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
darkwing, line 9: Dependabot used to have `live` updates but they've been removed with no promise of returning (https://github.community/t/dependabot-update-frequency-live-option/116321).

.github/dependabot.yml 
```diff
@@ -0,0 +1,12 @@
...
+      time: "06:00"
```
darkwing, line 10: This still seems helpful to me -- it would hit early-ish in everyone's day, so they can be reviewed and merged early, and quickly pulled the same day if we discover an issue not found by tests.

Gudahtt: Agreed!

.github/dependabot.yml 
```diff
@@ -0,0 +1,13 @@
...
+    target-branch: "main"
+    versioning-strategy: "increase-if-necessary"
+    open-pull-requests-limit: 10
```
darkwing, line 13: Setting to 10 initially because I envision we'll be getting a bunch of them upon first implementation, especially for the extension.

Gudahtt (approved):
LGTM! Seems like a good default for our libraries.

PR #49, opened 2021-04-09
Update prepublishOnly script

rekmarks:
Updates the `prepublishOnly` script to build, lint, and then test.

The reason we build first is because if we ever have e.g. JavaScript tests that are supposed to test things in `dist`, linting will fail if the build isn't up-to-date.


Gudahtt (approved):
LGTM!

PR #52, opened 2021-04-21
Format Markdown and YAML with Prettier

rekmarks:
Prettier can format Markdown and YAML (see [here](https://prettier.io/docs/en/index.html)), and I can't think of a reason not to leverage that here.


danfinlay (approved):
I like the change made on the readme on this example, so why not?

PR #54, opened 2021-04-29
Remove ts-node

rekmarks:
I don't know why we had this as a dev dependency. It's not referenced anywhere.


ryanml (approved):
let's get this dead code outta here!

PR #55, opened 2021-04-29
Update action workflow

rekmarks:
- Adds Node 16 to test matrix
- Adds reassuring `echo "Great success!"` to `all-tests-pass`


Gudahtt (approved):
LGTM!

PR #56, opened 2021-04-29
Only use dependabot for metamask updates

darkwing:
As suggested by @rekmarks , let's only use dependabot for MetaMask updates.


rekmarks (approved):
yaaasss

PR #57, opened 2021-04-29
Compile TypeScript modules to CommonJS instead of ES6

rekmarks:
This way, since `esModuleInterop` is `true`, our compiled `.js` files can be imported using both `require` and `import` without issue.


Gudahtt (approved):
LGTM!

PR #59, opened 2021-05-10
Add `@lavamoat/allow-scripts`

### LINKED ISSUE #51: build-test has ignore-scripts but no allow-scripts
the yarn install uses `--ignore-scripts`, we should tell users to use `@lavamoat/allow-scripts` to lock down scripts.

trying to build a project with @lavamoat/allow-scripts + scripts fails since it has `--ignore-scripts` but never calls allow-scripts.

https://github.com/MetaMask/metamask-module-template/blob/main/.github/workflows/build-test.yml#L21

possible solutions:

- we could add `yarn allow-scripts` to the build-test.yml
- we could add `yarn setup`, which is our convention for `yarn install --ignore-scripts && yarn allow-scripts`
- we could remove `--ignore-scripts` from our template and let the new repo edit the `build-test.yml` to suit its needs

Gudahtt:
The template now uses `@lavamoat/allow-scripts` to control which scripts run during install. The `setup` script has been added as well, which is our conventional name for the script that runs `install` followed by `allow-scripts`. Also a `.yarnrc` file has been added to ensure that `--ignore-scripts` is always set.

Closes #51

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -18,7 +18,8 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: yarn --frozen-lockfile --ignore-scripts
+      - run: yarn --frozen-lockfile
+      - run: yarn allow-scripts
       - run: yarn build
       - run: yarn lint
       - run: yarn test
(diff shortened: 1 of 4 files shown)


ryanml (approved):
++

PR #62, opened 2021-05-13
Clean the output directory on builds

rekmarks:
`tsc` will overwrite any existing files, but it will not delete anything that's not a subset of its output. Therefore, we should manually clean the `dist` directory on builds. I'd bet money that we have published nonsense files because of this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "lint:misc": "prettier '**/*.json' '**/*.md' '**/*.yml' --single-quote --ignore-path .gitignore",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
-    "build": "tsc --project ."
+    "build": "mkdir -p dist && rm -rf dist/* && tsc --project ."
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^1.0.5",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "lint:misc": "prettier '**/*.json' '**/*.md' '**/*.yml' --single-quote --ignore-path .gitignore",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
-    "build": "tsc --project ."
+    "build": "mkdir -p dist && rm -rf dist/* && tsc --project ."
```
Gudahtt, line 26: This will fail on Windows. Ideally we'd use something that works cross platform.

Gudahtt:
Are we sure we want to do this on every build? We could make this part of `prepublishOnly` instead.

This would prevent incremental builds for example, which might be nice during development.

rekmarks:
@Gudahtt that's a good point. Re: Windows compat, we'd either have to ship some kind of script or add `mkdirp` and `rimraf` as dev dependencies. Thoughts?

rekmarks:
@Gudahtt I managed to do it with only `rimraf`, which was already in the lockfile.

Gudahtt (approved):
LGTM!

PR #63, opened 2021-05-16
Adding .gitattributes file

ryanml:
Related: https://github.com/MetaMask/MetaMask-planning/issues/32

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto
+
+yarn.lock linguist-generated=false


.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto
```
ryanml, line 1: Welcoming suggestions for what else we may want in here, but this should be a good baseline

ryanml (commented):
Welcoming suggestions for what else we may want in here, but this should be a good baseline

Gudahtt (approved):
LGTM!

PR #64, opened 2021-05-17
Collect coverage from all source files

rekmarks:
Ensures that coverage is collected from all source files as opposed to just those that are tested. This behavior is controlled by the `collectCoverageFrom` field in `jest.config.js`.


Gudahtt (approved):
LGTM!

PR #65, opened 2021-05-18
Add .prettierrc.yml, don't run Prettier on CHANGELOG.md

rekmarks:
`@metamask/eslint-config` handles the Prettier configuration for `eslint-plugin-prettier`, but not for the use of `prettier` in e.g. the `lint:misc` package script.

This also updates the `lint:misc` script to exclude `CHANGELOG.md` from Prettier linting. This is because we're going to start auto-generating (and validating) changelogs using https://github.com/MetaMask/auto-changelog.

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,0 +1,4 @@
+quoteProps: consistent
+singleQuote: true
+trailingComma: all
+tabWidth: 2
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,6 +1,0 @@
-# All of these are defaults except singleQuote, but we specify them
-# for explicitness
-quoteProps: as-needed
-singleQuote: true
-tabWidth: 2
-trailingComma: all
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,4 +1,6 @@
-quoteProps: consistent
+# All of these are defaults except singleQuote, but we specify them
+# for explicitness
+quoteProps: as-needed
 singleQuote: true
 tabWidth: 2
 trailingComma: all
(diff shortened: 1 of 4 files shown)


.prettierrc.yml
```diff
@@ -0,0 +1,4 @@
+quoteProps: consistent
```
Gudahtt, line 1: :eyes: Did we ever remove this rule in the ESLint config? I had thought there was a general dislike of this on the team.

rekmarks: So far, no rejoinders have followed this: https://github.com/MetaMask/eslint-config/issues/175#issuecomment-828788208

rekmarks: We're doing `as-needed`.

Gudahtt (commented):
Could we use JavaScript rather than YAML for the configuration file? Of the four options Prettier supports, JavaScript is the most familiar for our team, and it supports comments (unlike JSON).

rekmarks:
@Gudahtt, @shanejonas has argued that we ought to use something other than `.js` for configs due to better editor support for e.g. JSON and YAML schemas. Wrong place to make that call, so I converted to `.js` in 2215020.

Gudahtt (approved):
LGTM!

Gudahtt:
>we ought to use something other than .js for configs due to better editor support for e.g. JSON and YAML schemas

Interesting :thinking: Hadn't considered this before.

I guess TOML is an option too. But I'm not against using YAML. I just don't prefer it because it's a confusing spec. And my initial reaction to seeing the diff was "Hey, those values aren't quoted!!" before realizing that it was YAML which doesn't need quotes.

PR #70, opened 2021-06-03
Fix release instructions

Gudahtt:
Step 4 of the release instructions still referenced the old main branch name (`master`). It has been updated to `main`. Also the branch names are now always in backticks.


NiranjanaBinoy (approved):
LGTM

PR #71, opened 2021-06-10
Add `@metamask/auto-changelog`

Gudahtt:
A basic empty CHANGELOG.md has been added to the project, along with `@metamask/auto-changelog` to help validate and keep the changelog up-to-date.

A step has been added to the "Build, Lint, and Test" workflow to validate the changelog.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -34,6 +34,12 @@ jobs:
       - run: yarn build
       - run: yarn lint
       - run: yarn test
+      - name: Validate RC changelog
+        if: ${{ startsWith(github.ref, 'release-v') }}
+        run: yarn auto-changelog validate --rc
+      - name: Validate changelog
+        if: ${{ !startsWith(github.ref, 'release-v') }}
+        run: yarn auto-changelog validate
   all-jobs-pass:
     name: All jobs pass
     runs-on: ubuntu-20.04
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,10 +35,10 @@ jobs:
       - run: yarn lint
       - run: yarn test
       - name: Validate RC changelog
-        if: ${{ startsWith(github.ref, 'release-v') }}
+        if: ${{ startsWith(github.ref, 'release/') }}
         run: yarn auto-changelog validate --rc
       - name: Validate changelog
-        if: ${{ !startsWith(github.ref, 'release-v') }}
+        if: ${{ !startsWith(github.ref, 'release/') }}
         run: yarn auto-changelog validate
   all-jobs-pass:
     name: All jobs pass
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       - run: yarn build
       - run: yarn lint
       - run: yarn test
+      - name: Validate RC changelog
+        if: ${{ startsWith(github.ref, 'release-v') }}
```
Gudahtt, line 38: Should probably change this :sweat_smile: 

Or maybe that can wait for the PR that adds the release automation

rekmarks: I think the release automation actions perform this validation, so maybe this step is unnecessary?

Gudahtt: Huh. I thought we had actions for creating a release PR and releasing upon the merge of that PR. This step is to validate the release PR before it's merged. I didn't know that was covered in either of the release actions.

rekmarks: Well, `create-release-pr` updates the changelog, and it seems to error if the changelog is invalid as well as `validateChangelog`: https://github.com/MetaMask/action-create-release-pr/blob/c220550e1faca0618d22b755dc8b1b4474473f17/src/package-operations.ts#L186

rekmarks: Hm, actually maybe not. That's certainly easy to add in that Action though. Not sure if we prefer it one place over the other? Here is fine, we should just update the prefix to `release/`.

Gudahtt: But that doesn't get run on subsequent commits on the release PR though. This is run on all release PR commits, and the `--rc` flag additionally ensures the manual categorization is done. It doesn't sound like it does the same thing at all :thinking:

rekmarks: I was referring to RC validation specifically. If we did that in `create-release-pr`, it would obviate the need for knowing the release branch prefix here. But, I'm agnostic!

rekmarks:
```suggestion
        if: ${{ startsWith(github.ref, 'release/') }}
```

.github/workflows/build-test.yml
```diff
@@ -34,6 +34,12 @@ jobs:
...
+        run: yarn auto-changelog validate --rc
+      - name: Validate changelog
+        if: ${{ !startsWith(github.ref, 'release-v') }}
```
rekmarks, line 41:
```suggestion
        if: ${{ !startsWith(github.ref, 'release/') }}
```

.github/workflows/build-test.yml
Gudahtt: Right, it still doesn't sound like we're on the same page about what this actually does though. `create-release-pr` _can't_ do that, because `--rc` here also ensures that changes have been categorized. That has to be done manually _after_ the release PR has been created.

rekmarks: Ah, I didn't realize it checks categorization. Thanks for clarifying!

rekmarks (approved):
LGTM!

PR #75, opened 2021-06-29
Add `prettier-plugin-packagejson`

Gudahtt:
This Prettier plugin sorts the properties in `package.json`. They're sorted using the library `sort-package-json`, which essentially sorts known keys by the order they appear in the npm documentation, and sorts alphabetically otherwise. The full details of the sorting order can be found here:
https://github.com/keithamus/sort-package-json/blob/master/defaultRules.md


rekmarks (approved):
LGTM!

PR #92, opened 2021-12-19
Add git working tree check to CI

rekmarks:
Adds a `! git diff --exit-code` check to CI after the build step to ensure that the working tree remains clean after building. We don't expect to _need_ this check for every repository, but it should always pass for every repository, and is very fast as well!

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -32,6 +32,13 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn allow-scripts
       - run: yarn build
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty after building"
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -32,6 +32,13 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn allow-scripts
       - run: yarn build
+      - name: Require clean working directory
```
Gudahtt, line 35: Should this go last? In case the other commands result in a dirty working tree as well. They shouldn't, but I can see how it might happen with certain test setups for example, e.g. coverage files.

Gudahtt (dismissed):
LGTM! Left one suggestion but this is still a good idea.

.github/workflows/build-test.yml
rekmarks: Yeah, this is a good idea. Added!

Gudahtt (approved):

PR #93, opened 2022-01-11
Prevent prettier error on unmatched patterns

rekmarks:
Add the `--no-error-on-unmatched-pattern` flag to the `prettier` invocation to prevent it from erroring if a particular file pattern is not matched. The patterns we have may be unmatched in certain repositories, for example in monorepos without `.yml` files in a particular package.

The reason we want to ignore these errors rather than remove the unmatched patterns is because it's very easy to forget to add them back should matching files be added, and then those files will not be linted for a potentially long period of time.


mcmire (approved):
Makes sense.

PR #97, opened 2022-04-04
Upgrade Jest and use TS for Jest config

mcmire:
v27 is the latest version of Jest and introduces these major changes:

* `node` is now the default environment instead of `jsdom`. This
  results in faster runtimes.
* The runner has been completely rewritten as `jest-circus` which is
  also the default, replacing `jest-jasmine2`.
  * This should have no impact on us as we don't rely on any
    Jasmine-specific APIs, but is nice to know.
* The "modern" version of Fake Timers (enabled via
  `jest.useFakeTimers("modern")`) is now the default.

More info here: <https://jestjs.io/blog/2021/05/25/jest-27>

Additionally, to ensure that we make use of all of the latest and
greatest, this PR regenerates the Jest config file (as
`jest.config.ts`). This has the added improvement that each option is
now commented so we know exactly what each one does.


jest.config.ts
```diff
@@ -0,0 +1,194 @@
...
+  // automock: false,
+
+  // Stop running tests after `n` failures
+  // bail: 0,
+
+  // The directory where Jest should store its cached dependency information
+  // cacheDirectory: "/private/var/folders/fk/c3y07g0576j8_2s9m01pk4qw0000gn/T/jest_dx",
+
+  // Automatically clear mock calls, instances and results before every test
+  clearMocks: true,
```
Gudahtt, line 17: Why has `resetMocks` and `restoreMocks` been replaced by `clearMocks`? `clearMocks` does a subset of what `resetMocks` does, so it seems like the worse choice. We need both `resetMocks` and `restoreMocks` to ensure no mocks or mock state is shared between tests.

jest.config.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  // Indicates whether the coverage information should be collected while executing the test
+  collectCoverage: true,
+
+  // An array of glob patterns indicating a set of files for which coverage information should be collected
+  // collectCoverageFrom: undefined,
```
Gudahtt, line 23: Removing this seems like a regression as well. Without this, we get no signal from files that are entirely untested.

jest.config.ts
```diff
@@ -0,0 +1,194 @@
...
+  // A list of reporter names that Jest uses when writing coverage reports
+  // coverageReporters: [
+  //   "json",
+  //   "text",
+  //   "lcov",
+  //   "clover"
+  // ],
+
+  // An object that configures minimum threshold enforcement for coverage results
+  // coverageThreshold: undefined,
```
Gudahtt, line 45: Unclear why coverage thresholds were removed

Gudahtt:
I started going through the new config, but it looks like most of our old config was discarded. Perhaps by mistake? Also lots of lint errors.

jest.config.ts
mcmire: Ah yes. Good catch. I'll add these back.

mcmire:
@Gudahtt Apologies for the rush job on this, just wanted to make a PR based on the changes I'd made to `eth-json-rpc-infura` before I forgot. Should be better now.

jest.config.ts
```diff
@@ -0,0 +1,204 @@
...
+  // testEnvironment: "jest-environment-node",
+
+  // Options that will be passed to the testEnvironment
+  // testEnvironmentOptions: {},
+
+  // Adds a location field to test results
+  // testLocationInResults: false,
+
+  // The glob patterns Jest uses to detect test files
+  // testMatch: [
```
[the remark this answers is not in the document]
Gudahtt: I believe it was because it _used_ to diverge from the standard. The default seems fine to me.

.eslintrc.js
```diff
@@ -21,6 +21,14 @@ module.exports = {
       files: ['*.test.ts', '*.test.js'],
       extends: ['@metamask/eslint-config-jest'],
     },
+
+    {
+      files: ['jest.config.ts'],
+      rules: {
+        // TODO: Migrate to our shared config
+        'import/no-anonymous-default-export': 'off',
```
Gudahtt, line 29: :eyes: 

I was curious what this rule was for so I just looked up the original PR where it was added. Apparently the advantage is that it "improves the greppability of the codebase". That is, when searching for an identifier, sometimes it can be hard to find what you're looking for if that identifier is only in the filename and not represented in the file itself.

Gudahtt: I have definitely had that problem myself, so in general this rule seems beneficial to me. But it does seem out-of-place when applied to a top-level config file.

Gudahtt (approved):
LGTM!

PR #98, opened 2022-04-06
Typecheck Jest config

mcmire:
The Jest documentation includes a [TypeScript example][1] that
suggests how a user-provided config object can be typechecked. This
commit incorporates these suggestions. Because we assign the default
export to a variable first, this obviates the need to turn off the
ESLint `import/no-anonymous-default-export` rule.

[1]: https://jestjs.io/docs/configuration


Gudahtt (approved):
LGTM!

PR #102, opened 2022-04-25
Extract build-only TypeScript config

mcmire:
Currently, test files are excluded from TypeScript's purview via the
`exclude` option in `tsconfig.json`. This is helpful when compiling
because it means that test files will not be included in the build
directory. However, for development, this is unnecessary, as we want
TypeScript to "see" and resolve types across all files in the project.

This commit splits off the build-specific settings from `tsconfig.json`
into a `tsconfig.build.json` file, which will exclusively be used in the
`yarn build` script.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,12 +1,11 @@
 {
   "compilerOptions": {
-    "declaration": true,
     "esModuleInterop": true,
     "inlineSources": true,
     "lib": ["ES2020"],
     "module": "CommonJS",
     "moduleResolution": "node",
-    "outDir": "dist",
+    "noEmit": true,
     "sourceMap": true,
     "strict": true,
     "target": "es2017"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,5 +8,5 @@
     "strict": true,
     "target": "es2017"
   },
-  "exclude": ["./dist/**"]
+  "exclude": ["./dist/**/*"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,13 +5,10 @@
     "inlineSources": true,
     "lib": ["ES2020"],
     "module": "CommonJS",
-    "moduleResolution": "Node",
+    "moduleResolution": "node",
     "outDir": "dist",
-    "rootDir": "src",
     "sourceMap": true,
     "strict": true,
-    "target": "ES2017"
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -1,17 +1,13 @@
...
-    "moduleResolution": "Node",
-    "outDir": "dist",
-    "rootDir": "src",
+    "moduleResolution": "node",
+    "noEmit": true,
     "sourceMap": true,
     "strict": true,
-    "target": "ES2017"
-  },
-  "exclude": ["./src/**/*.test.ts"],
```
Gudahtt, line 15: Maybe worth either preserving a broader `include` rule, or adding an `exclude` for `dist` here, for the reasons discussed here: https://github.com/MetaMask/eth-block-tracker/pull/103#discussion_r858878740

tsconfig.json
```diff
@@ -1,17 +1,13 @@
 {
   "compilerOptions": {
-    "declaration": true,
     "esModuleInterop": true,
     "inlineSources": true,
```
Gudahtt, line 4: Since this is set to `noEmit`, do these two sourcemap settings do anything? If so maybe we should disable them.

mcmire: You're right. Changed in [375755d](https://github.com/MetaMask/metamask-module-template/pull/102/commits/375755dce94bb1d63c138c7b433eb2a4df68eb95).

tsconfig.json
```diff
@@ -1,17 +1,12 @@
...
-    "sourceMap": true,
+    "moduleResolution": "node",
+    "noEmit": true,
     "strict": true,
-    "target": "ES2017"
+    "target": "es2017"
   },
-  "exclude": ["./src/**/*.test.ts"],
-  "include": ["./src/**/*.ts"]
+  "exclude": ["./dist/**"]
```
Gudahtt, line 11: In the typescript docs, usually to exclude or include an entire directory recursively, this pattern is used:

```suggestion
  "exclude": ["./dist/**/*"]
```

Though I am uncertain whether these are evaluated differently or not :thinking:

mcmire: I am uncertain as well! But I will make this change just to be safe.

Gudahtt (approved):
LGTM!

PR #106, opened 2022-06-07
Add jest-it-up

rekmarks:
Adds `jest-it-up` to the template. [`jest-it-up`](https://github.com/rbardini/jest-it-up) is a neat package that detects when coverage has increased, and automatically your coverage minimums in the Jest config. The only requirements made of your Jest config are that (1) it is a conventional `jest.config.js` file and (2) its `coverageReporters` includes `json-summary`. Consequently, this PR converts the Jest config file from TypeScript back to plain JavaScript, and adds the required coverage reporter.

I believe that the benefits conferred by `jest-it-up` outweigh those of using TypeScript for the config file, which rarely changes and is never consumed by application code. Meanwhile,  `json-summary` outputs a single JSON file to the `coverage/` directory and does not appear to impact performance.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -34,25 +34,17 @@ jobs:
       - run: yarn build
       - run: yarn lint
       - run: yarn test
-      # Intended to catch any coverage-related changes
-      - name: Require clean working directory after testing
-        shell: bash
-        run: |
-          if ! git diff --exit-code; then
-            echo "Working tree dirty after testing"
-            exit 1
-          fi
       - name: Validate RC changelog
         if: ${{ startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate --rc
       - name: Validate changelog
         if: ${{ !startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate
-      - name: Require clean working directory after building
+      - name: Require clean working directory
         shell: bash
         run: |
           if ! git diff --exit-code; then
-            echo "Working tree dirty after building"
+            echo "Working tree dirty at the end of CI job"
             exit 1
           fi
   all-jobs-pass:
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -34,13 +34,21 @@ jobs:
       - run: yarn build
       - run: yarn lint
       - run: yarn test
+      # Intended to catch any coverage-related changes
+      - name: Require clean working directory after testing
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty after testing"
+            exit 1
+          fi
       - name: Validate RC changelog
         if: ${{ startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate --rc
       - name: Validate changelog
         if: ${{ !startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate
-      - name: Require clean working directory
+      - name: Require clean working directory after building
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "prepublishOnly": "yarn build:clean && yarn lint && yarn test",
     "setup": "yarn install && yarn allow-scripts",
     "test": "jest",
+    "posttest": "jest-it-up",
```
FrederikBolding, line 24: note: `posttest` is not automatically run in Yarn v3, so this needs to be added if this template starts using Yarn v3.

.github/workflows/build-test.yml
```diff
@@ -34,13 +34,21 @@ jobs:
...
+            exit 1
+          fi
       - name: Validate RC changelog
         if: ${{ startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate --rc
       - name: Validate changelog
         if: ${{ !startsWith(github.head_ref, 'release/') }}
         run: yarn auto-changelog validate
-      - name: Require clean working directory
+      - name: Require clean working directory after building
```
FrederikBolding, line 51: Why does the template run this so long after building? This doesn't match the skunkworks behaviour at least.

FrederikBolding (approved):
lgtm - left a few comments though

jest.config.js
```diff
@@ -3,9 +3,7 @@
  * https://jestjs.io/docs/configuration
  */
 
-import type { Config } from '@jest/types';
-
-const config: Config.InitialOptions = {
+module.exports = {
```
mcmire, line 6: Yeah this is fine to convert this back. I don't think the types were giving us that much information anyway.

package.json
mcmire: Yeah, we should proactively work out a solution here. Why use a `post*` hook here at all? In the `test` script can we say `jest && jest-it-up`?

Gudahtt:
The template starts with required coverage at 100%, so I didn't think it was needed here. No need to ratchet up coverage if it's already at max. I'm not opposed to adding it though if you think it'd be useful.

mcmire:
@Gudahtt Ah sorry, I suggested that Erik add something to the module template because of [this change](https://github.com/MetaMask/providers/pull/209/files#diff-5c9b321fff56771d69adf06a25e54768beec0733c826f03915b7211b792b93abR24) in `providers`. It didn't occur to me that [this line](https://github.com/MetaMask/providers/pull/209/files#diff-5c9b321fff56771d69adf06a25e54768beec0733c826f03915b7211b792b93abR33) was the real add, and that it was borrowed from `snaps-skunkworks` where `jest-it-up` is already a thing. For new projects going forward we may not want `jest-it-up` but in the interest of keeping existing projects standardized I don't mind this change myself.

package.json
rekmarks: Hm, yeah, that's better.

rekmarks: 531110d

.github/workflows/build-test.yml
rekmarks: Ohhh, I was not paying attention when I wrote this. 🤦 

Basically it's putting the dirty check at the end of whole CI pipeline. That's probably fine for most cases.

Moved back to a single check here: db588cc

rekmarks:
@Gudahtt @mcmire I think adding `jest-it-up` to the template would be useful because (1) we use the template to perform "repo standardization" tasks on existing repos, and (2) although the template requires requires 100%, we need not trust the better angels of our nature alone to maintain that.

mcmire (dismissed):
LGTM!

mcmire (approved):

PR #108, opened 2022-07-07
Upgrade to Yarn v3

mcmire:
The package manager used has been upgraded from Yarn v1 to Yarn v3. All
of the steps in the [migration guide][1] have been followed, so
effectively everything should work the same way it did before. (Note
that we are not using Plug'n'Play because it seems to prevent TypeScript
from finding type definitions via `@types/*` packages.)

The `setup` script has been replaced with a Yarn v3 plugin that will
automatically run `allow-scripts` after install. This means that we can
now use `yarn install` again as normal. This change was made to address
a problem that happened when running `yarn setup` for the first time.
Yarn 3 doesn't automatically install itself until you run the `install`
command directly, so running the `setup` script before `install` would
have thrown an error. This is non-obvious, so we chose to remove `setup`
altogether.

See [original source in `controllers`][2].

[1]: https://yarnpkg.com/getting-started/migration
[2]: https://github.com/MetaMask/controllers/commit/d92bec08ec4f1994097e36b69c07bfe54c87270c


Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

mcmire:
🙏🏻

PR #110, opened 2022-08-05
add most common ways of setting node version

BelfordZ:
We leave nvmrc as the default for releases, but we add config files for the most popular 3 ways of setting a node version.

It also updates setup-node to use v3, which now has [`node-version-file`](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file)


README.md
```diff
@@ -22,8 +22,11 @@ _Add examples here_
 
 ### Setup
 
-- Install [Node.js](https://nodejs.org) version 12
-  - If you are using [nvm](https://github.com/creationix/nvm#installation) (recommended) running `nvm use` will automatically choose the right node version for you.
+- Install [Node.js](https://nodejs.org) using one of the following:
+  - (recommended) If you are using [nvm](https://github.com/creationix/nvm#installation) running `nvm use` will [automatically](./nvmrc) choose the right node version for you.
```
BelfordZ, line 26:
```suggestion
  - *(recommended)* If you are using [nvm](https://github.com/creationix/nvm#installation) running `nvm use` will [automatically](./nvmrc) choose the right node version for you.
```

mcmire:
Hey @BelfordZ, thanks for making these changes. I like the update to `@actions/setup-node`, that definitely makes sense. As for adding support for other Node managers, however, I'm of two minds. On one hand, it's very considerate for non-MetaMask developers, in case someone out there who is not using `nvm` wants to contribute to one of our libraries. On the other hand, it's potentially problematic for MetaMask developers. If we allow for multiple Node managers it creates a possible fracture within MetaMask, where some developers are setting up their development environment one way and some are doing so another way. From a developer experience perspective, this is sort of a nightmare. I believe we should strive for consistency across our development teams, because it means that it's easier for everyone to help each other out in case there are issues. Once a developer is allowed to have another way of following a process it can create frustration because if you have to change that process, you have to account for more possibilities. I know that `nvm` has its problems, it's definitely not the best tool to use, but right now everyone uses it, and I think it's best we keep it that. For this reason I'm not so keen on supporting `nodenv` and `asdf` as well. That said I'm curious to know what other people think.

Gudahtt:
^ Agreed completely!

Open to discussing supporting additional tools or switching tools, but if there isn't a clear case for how it benefits our team, I'd rather avoid the clutter.

Gudahtt:
I'll close this for now; feel free to reopen the discussion if you want.

PR #111, opened 2022-08-09
Test on Node 18

rekmarks:
Starts testing the module template on Node 18 in CI, and updates dependencies for Node 18 compatibility.


package.json
```diff
@@ -37,7 +37,7 @@
     "eslint-config-prettier": "^8.1.0",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jest": "^24.3.4",
-    "eslint-plugin-jsdoc": "^36.1.0",
+    "eslint-plugin-jsdoc": "^39.2.9",
```
rekmarks, line 40: This is the minimum version of the package that's compatible with Node 18. They make no note of this change in their changelog.

package.json
```diff
@@ -29,23 +29,23 @@
...
+    "@types/jest": "^28.1.6",
     "@types/node": "^17.0.23",
     "@typescript-eslint/eslint-plugin": "^4.21.0",
     "@typescript-eslint/parser": "^4.21.0",
     "eslint": "^7.23.0",
     "eslint-config-prettier": "^8.1.0",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jest": "^24.3.4",
-    "eslint-plugin-jsdoc": "^36.1.0",
+    "eslint-plugin-jsdoc": "^39.2.9",
```
Gudahtt, line 40: This is outside the range requested by our ESLint config: https://github.com/MetaMask/eslint-config/blob/main/packages/base/package.json#L41

Ideally we'd update our config first

package.json
```diff
@@ -29,23 +29,23 @@
...
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.3.1",
-    "jest": "^27.5.1",
+    "jest": "^28.1.3",
     "jest-it-up": "^2.0.2",
     "prettier": "^2.2.1",
     "prettier-plugin-packagejson": "^2.2.17",
     "rimraf": "^3.0.2",
-    "ts-jest": "^27.1.4",
+    "ts-jest": "^28.0.7",
```
Gudahtt, line 48: This version of `ts-jest` requires a minimum TypeScript version of v4.3, but we're still on v4.2

package.json
Gudahtt: This is done in #114

package.json
Gudahtt: This was updated in #116

mcmire (approved):
LGTM!

Gudahtt (approved):
LGTM!

PR #118, opened 2022-10-05
Remove require-additional-reviewer workflow

mcmire:
The require-additional-reviewer workflow was added in order to ensure that release PRs received an extra check before they were merged. Now that we've added automated NPM publishing — which includes an approval step of its own — this action is no longer necessary. Besides, it's never worked right on forks.


Gudahtt:
The npm publish approval doesn't really serve the same purpose. By the time a release is published to GItHub, it's difficult to change it.

mcmire:
@Gudahtt Right, hence the approval step — you can always stop the publish from occurring if you really don't want to. Right?

Gudahtt:
The npm approval step happens _after_ the release is published to GitHub, but before the npm publishing step. It's after the PR is merged.

`require-additional-reviewer` is meant to ensure our _regular_ standards for PR review are upheld because the bot authors it, so we don't get the built-in exclusion of the author as reviewer.

mcmire:
@Gudahtt Ah, okay. I must have misunderstood the discussion in Slack about this then. I'll close this.

PR #121, opened 2022-10-07
Remove Dependabot

FrederikBolding:
Since Dependabot doesn't support Yarn 3 and the module template uses Yarn 3, we should not be recommending it for new modules that are built using the current module template.

I believe we might add Renovate as an alternative, but for now, that work is still pending.


Gudahtt (approved):
LGTM!

PR #122, opened 2022-10-07
Add `SKIP_PREPACK` environment variable to publish actions

Mrtenz:
This adds the missing `SKIP_PREPACK` environment variable to the publish actions, to avoid running scripts.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -62,6 +62,8 @@ jobs:
       - name: Dry Run Publish
         # omit npm-token token to perform dry run publish
         uses: MetaMask/action-npm-publish@v1.1.0
+        env:
+          SKIP_PREPACK: true
 
   publish-npm:
     environment: npm-publish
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
LGTM!

To add context for others: this was copied from the controllers repo. The current publish action is broken because it will try to run `prepublishOnly` despite there being no `node_modules` directory. This is unnecessary because the build has finished already by this point.

mcmire:
@Gudahtt Got it. So... do we even need the `prepack` script, then? My understanding was that it was added for the publish step.

mcmire (approved):

Gudahtt:
@mcmire `prepack` is the replacement for `prepublishOnly` (which Yarn v3 doesn't have). It's to ensure that if we manually publish, we don't forget to build first. This PR is intended to skip that here in CI because we have definitely already built.

We have discovered that this is incomplete; we need to add another script that uses this SKIP_PREPACK flag

Gudahtt (approved):
LGTM!

PR #123, opened 2022-10-07
Publish docs as part of release workflow

Mrtenz:
The docs workflow can't be triggered by merging the release commit, because the default `GITHUB_TOKEN` does not permit that. This adds the release docs workflow to the `publish-release` workflow, as follows:

- The staging docs are published when the `publish-release` succeeds, i.e., after merging a publish PR.
- The latest and versioned docs are published when the `publish-npm` succeeds, i.e., after the publish is approved.


.github/workflows/publish-release.yml
```diff
@@ -88,3 +113,21 @@ jobs:
           npm-token: ${{ secrets.NPM_TOKEN }}
         env:
           SKIP_PREPACK: true
+
+  publish-release-to-gh-pages:
+    needs: publish-npm
+    name: Publish docs to `${{ github.ref_name }}` directory of `gh-pages` branch
```
Gudahtt, line 119: `github.ref_name` is usually set according to the event that triggered the workflow. In this case, that event won't have the correct refname set because it hasn't been tagged yet at that point. We'll; have to get the version name from somewhere, like `package.json` maybe.

.github/workflows/publish-release.yml
```diff
@@ -76,7 +99,9 @@ jobs:
       - uses: actions/cache@v3
         id: restore-build
         with:
-          path: ./dist
+          path: |
+            ./dist
+            ./docs
```
Gudahtt, line 104: Are the docs needed here? We intentionally kept the cache minimal for the publishing step.

.github/workflows/publish-release.yml
Gudahtt: Ricky has a script for this already: https://github.com/MetaMask/action-publish-release/blob/a0e96f152a64664437d4dc5bfc381f42fa75978a/action.yml#L12-L19

.github/workflows/publish-release.yml
```diff
@@ -39,12 +39,33 @@ jobs:
...
           key: ${{ github.sha }}
 
+  publish-staging-to-gh-pages:
```
Gudahtt, line 51: It's worth commenting here on why this step is being duplicated between here and the normal staging workflow

e.g.

```suggestion
  # Normally this happens during the `publish-main-docs` workflow.
  # That workflow will not run in this case because GitHub Actions workflows cannot be
  # triggered by events that were themselves triggered by a GitHub Actions workflow.
  publish-staging-to-gh-pages:
```

Gudahtt: Actually... wait.... this isn't true. The staging publishing worked just fine for `abi-utils`, because the RC gets merged by a human.

We only need the `latest` and tag docs publishing step here. It's just the tag trigger that was broken, because the tagging is solely done by GitHub actions.

We can remove this then.

.github/workflows/publish-release.yml
Mrtenz: Since we can remove the staging deployment, this is definitely not needed anymore. I wasn't sure if the cache would be restored if the `path` was different, but it's irrelevant now.

.github/workflows/publish-release.yml
```diff
@@ -88,3 +88,33 @@ jobs:
...
+
+  publish-release-to-gh-pages:
+    runs-on: ubuntu-latest
+    needs: get-release-version
+    name: Publish docs to `${{ needs.get-release-version.outputs.RELEASE_VERSION }}` directory of `gh-pages` branch
+    permissions:
+      contents: write
+    uses: ./.github/workflows/publish-docs.yml
+    with:
+      destination_dir: ${{ needs.get-release-version.outputs.RELEASE_VERSION }}
```
Gudahtt, line 110: Elsewhere I see steps referenced via `steps.[step id].outputs`. e.g.:

```suggestion
    name: Publish docs to `${{ steps.get-release-version.outputs.RELEASE_VERSION }}` directory of `gh-pages` branch
    permissions:
      contents: write
    uses: ./.github/workflows/publish-docs.yml
    with:
      destination_dir: ${{ steps.get-release-version.outputs.RELEASE_VERSION }}
```

Does `needs` work as well? I've never used that before.

Gudahtt: Oh I see, the example I was looking at was for reusing the output _within_ a job. Here we're using it in a separate job. Different case.

Gudahtt, line 110: It'd be nice if this were a list, so we could deploy the same docs to two different places. Future enhancement maybe.

Gudahtt (approved):
LGTM!

PR #124, opened 2022-10-12
Remove invalid `runs-on` from publish workflow

Mrtenz:
`runs-on` is not compatible with `uses`/`with`:

```
The workflow is not valid. .github/workflows/publish-release.yml
(Line: 108, Col: 5): Unexpected value 'uses' .github/workflows/publish-release.yml
(Line: 109, Col: 5): Unexpected value 'with'
```

https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow

I added this on accident in #123.


Gudahtt (approved):
LGTM!

PR #126, opened 2022-10-12
Check out repository in `get-release-version` for publishing docs

Mrtenz:
`get-release-version` was missing a checkout, so running it would result in an error because the script cannot be found.


Gudahtt (approved):
LGTM!

PR #131, opened 2022-11-02
Upgrade to nodejs v18

legobeat:
* v12 is EOL since 2022-04-30
* v14 will reach EOL 2023-04-30
* v18 LTS 2022-10-25 - 2023-10-23

https://github.com/nodejs/release#release-schedule


.nvmrc
```diff
@@ -1 +1 @@
-v14
+v18
```
FrederikBolding, line 1: I think we should stick to Node 16 until the extension upgrades to 18.

legobeat: Fair enough; will address!

Gudahtt:
Do we want to stop supporting Node.js v14 in our libraries? Lots of them are used by other projects outside of MetaMask. We've been tracking the maintenance LTS schedule (roughly) as a courtesy to other users of our libraries. That schedule would have us supporting v14 until next year.

legobeat:
> Do we want to stop supporting Node.js v14 in our libraries? Lots of them are used by other projects outside of MetaMask. We've been tracking the maintenance LTS schedule (roughly) as a courtesy to other users of our libraries. That schedule would have us supporting v14 until next year.

With that in light, let's add 14 back to the ci matrix.

Gudahtt:
Sounds good! Totally open to dropping v14 if it benefits us enough to be worth the tradeoff of slightly inconveniencing other projects. Just wanted to highlight that as a tradeoff.

legobeat:
I'm agnostic. Either way, I think breaking existing expectations needs some prior announcement and should be a separate change.

legobeat:
Superceded by #136

PR #132, opened 2022-11-02
Node versioning

legobeat:
Normalize versioning and upgrade version matrix. Ported over changes from #110 without the extra tooling.


Gudahtt (approved):
LGTM! Thanks @legobeat and @BelfordZ, we should have used this years ago.

PR #136, opened 2022-11-02
Bump minimum Node.js version to 16

Mrtenz:
Node.js 14 will be EOL in a few months, and many of our projects (including the extension) require Node.js 16+ as well.


FrederikBolding (approved):
We may even be able to bump to 18 soon. Testing updating the extension right now.

PR #137, opened 2022-11-02
Replace deprecated GitHub actions commands

Mrtenz:
`::set-output` is deprecated. This replaces it with the updated way to do it.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -21,16 +21,16 @@ jobs:
         uses: actions/checkout@v3
       - name: Get Node.js version
         id: nvm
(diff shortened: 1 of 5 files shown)


FrederikBolding (approved):

.github/workflows/publish-docs.yml
```diff
@@ -21,16 +21,16 @@ jobs:
         uses: actions/checkout@v3
       - name: Get Node.js version
         id: nvm
-        run: echo "::set-output name=NODE_VERSION::$(cat .nvmrc)"
+        run: echo "NODE_VERSION=$(cat .nvmrc)" >> "$GITHUB_OUTPUT"
```
legobeat, line 24: @Mrtenz Curious: Why was this preferred over #132 ?

PR #138, opened 2022-11-03
Support node 14

legobeat:
* Add back support for maintenance version 14 (see https://github.com/MetaMask/metamask-module-template/pull/131#issuecomment-1300362039)
* CI build for current nodejs version 19

(suggestion: we could change the policy to only support the latest maintenance release starting with the EOL of v14 in 2023-04, which would make the number of built versions more reasonable again)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -14,7 +14,7 @@ jobs:
       YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x, 19.x]
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x, 19.x]
```
Gudahtt, line 17: Hmm, we haven't typically tested non-LTS versions.  I guess because they sometimes have stability issues that we don't care about, because we aren't eager enough to try using them.

But if that happens in a particular package we can just stop testing it. This seems good and useful.

Gudahtt (approved):
LGTM!

.github/workflows/build-test.yml
legobeat: after having had this mulled for a bit: IMO it would make sense to have different default policies for library packages (down to oldest/latest maintenance) vs applications/services (lts enough). wdyt?

Gudahtt: That seems reasonable!

This template is intended for libraries in particular. We have few enough applications/services that we haven't created a template for them.

legobeat: makes sense!

PR #143, opened 2022-11-09
Run publish after build, lint, test completed successfully

Mrtenz:
This makes the publish workflow only run after the build, lint, and test steps have completed successfully. It does a workflow call to `publish-release.yml`, if the current commit is a release commit.


.github/workflows/build-test.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           key: yarn-cache-${{ runner.os }}-${{ steps.yarn-version.outputs.YARN_VERSION }}-${{ hashFiles('yarn.lock') }}-${{ matrix.node-version }}
       - name: Install Yarn dependencies
         run: yarn --immutable
+
```
Mrtenz, line 37: Since this file has grown a lot, I've added some spaces between jobs to make it more readable.

mcmire (approved):
Nice, LGTM!

Gudahtt (approved):
LGTM!

PR #144, opened 2022-11-10
Update action-npm-publish to support Yarn protocols

Gudahtt:
The action `MetaMask/action-npm-publish` has been updated to v2, which uses Yarn for publishing rather than npm. This ensures the publish step works correctly with Yarn-specific dependency protocols such as `workspace:`.

The `.yarn-state.yml` file needs to be cached and restored during publish to ensure that Yarn can run the `prepack` script. The prepack step is a no-op here because we have `SKIP_PREPACK` set, but there is no way to tell Yarn to not attempt running it.

Since we're no longer using npm, we don't need to set `ignore-scripts` in the npm config anymore either.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -26,7 +26,9 @@ jobs:
       - uses: actions/cache@v3
         id: restore-build
         with:
-          path: ./dist
+          path: |
+            ./dist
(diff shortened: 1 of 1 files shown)


Mrtenz (approved):

mcmire (approved):
Sweet!

PR #145, opened 2022-11-10
Use `actions/setup-node` cache

Mrtenz:
[`actions/setup-node` has an option to cache dependencies](https://github.com/actions/setup-node#caching-global-packages-data), which essentially does what we were doing manually, using. `actions/cache` internally. Using it means we can remove the manual caching step.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -9,9 +9,6 @@ jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
-    outputs:
-      YARN_CACHE_DIR: ${{ steps.yarn-cache-dir.outputs.YARN_CACHE_DIR }}
-      YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
     strategy:
       matrix:
         node-version: [14.x, 16.x, 18.x, 19.x]
@@ -21,17 +18,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
-      - name: Get Yarn cache directory
-        run: echo "YARN_CACHE_DIR=$(yarn config get cacheFolder)" >> "$GITHUB_OUTPUT"
-        id: yarn-cache-dir
-      - name: Get Yarn version
-        run: echo "YARN_VERSION=$(yarn --version)" >> "$GITHUB_OUTPUT"
-        id: yarn-version
-      - name: Cache Yarn dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.yarn-cache-dir.outputs.YARN_CACHE_DIR }}
-          key: yarn-cache-${{ runner.os }}-${{ steps.yarn-version.outputs.YARN_VERSION }}-${{ hashFiles('yarn.lock') }}-${{ matrix.node-version }}
+          cache: 'yarn2'
       - name: Install Yarn dependencies
         run: yarn --immutable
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -9,29 +9,13 @@ jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
-    outputs:
-      YARN_CACHE_DIR: ${{ steps.yarn-cache-dir.outputs.YARN_CACHE_DIR }}
-      YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
-    strategy:
-      matrix:
-        node-version: [14.x, 16.x, 18.x, 19.x]
```
legobeat, line 17: What's behind this change? node caches are in general not compatible between major versions.

https://github.com/actions/setup-node#matrix-testing

legobeat: (In either case, if the intention is to build a single version only, I guess that's significant enough to belong in a separate PR?)

Gudahtt: We're caching the Yarn cache directory here, which is a collection of .zip files that represent exactly what was downloaded from the package repository. I thought this cache would work the same across Node.js versions. I'm not certain though.

The only Node.js-version-specific steps I'm aware of are postinstall steps (e.g. especially for native addons that use the Node.js C/C++ API). So caching `node_modules` would definitely break in some cases between Node.js versions.

.github/workflows/build-test.yml
```diff
@@ -9,29 +9,13 @@ jobs:
   prepare:
```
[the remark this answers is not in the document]
Mrtenz: The point of it is to ensure that each job after it doesn't need to pull in the dependencies again. Build runs at the same time as lint and test.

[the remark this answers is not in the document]
mcmire: Ah, gotcha, makes sense.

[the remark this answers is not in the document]
Gudahtt: We should probably be using the `--immutable-cache` Yarn option to make that more clear, and alert us if the caching fails

.github/workflows/build-test.yml
mcmire: I disagree with this change as well. I think we should install Yarn dependencies for each Node version rather than for the development version in `.nvmrc` once. It may be that some of the dependencies in `package.json` are only compatible with the development version or higher, but any incompatibilities with versions less than the development version won't be caught in CI if we simply reuse the development Node version when installing dependencies.

Mrtenz: This should be caught by the other jobs, as they still run `yarn install` for each specific Node version.

Gudahtt:
It sounds like it might be more efficient to stick with `action/cache`. This issue indicates that `action/cache` is better at reusing the cache when the lockfile changes, so that only missing packages need to be downloaded: https://github.com/actions/setup-node/issues/325

.github/workflows/build-test.yml
mcmire: Alright, sounds good.

Mrtenz:
@Gudahtt How does that work? I thought that if the hash is different, the entire cache is invalidated.

Gudahtt:
Restore keys: https://github.com/actions/cache#inputs

Setup-node doesn't use them it sounds like. But we can to get cache hits for similar caches.

Gudahtt (approved):
LGTM!

tuanggo (commented):
open

PR #146, opened 2022-11-17
Add `dependabot.yml` again

Mrtenz:
We removed `dependabot.yml` in #121, because it was not supported by Yarn 3. It is now supported, meaning that we can add it again.


legobeat (approved):

Gudahtt (approved):
LGTM!

PR #147, opened 2022-11-17
Add `NPM_TOKEN` to publish-release workflow call

Mrtenz:
Secrets in a workflow called through `workflow_call` require secrets to be defined explicitly: https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -143,3 +143,5 @@ jobs:
     permissions:
       contents: write
     uses: ./.github/workflows/publish-release.yml
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -2,6 +2,9 @@ name: Publish Release
 
 on:
   workflow_call:
+    secrets:
+      NPM_TOKEN:
+        required: true
 
 jobs:
   publish-release:


Gudahtt (approved):
LGTM!

.github/workflows/build-test.yml
```diff
@@ -143,3 +143,5 @@ jobs:
     permissions:
       contents: write
     uses: ./.github/workflows/publish-release.yml
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
mcmire, line 147: @Mrtenz Will this work? The `NPM_TOKEN` secret is not available outside of the `npm-publish` environment, so I'm wondering whether this will be undefined initially.

Gudahtt: Oops. I think you're right. We'll need to move this into the `npm-publish` environment instead/as well.

Gudahtt: Oh, surprisingly, this _does_ work after all. I just tested it: https://github.com/Gudahtt/eslint-config-fork

The secret is carried through despite the top-level job not being in the environment. Maybe it only checks the environment in the job where the secret is directly read :thinking:

Gudahtt: I tried publishing that test package without these changes, just to confirm this was the cause of the issue, and sure enough it failed: https://github.com/Gudahtt/eslint-config-fork/actions/runs/3493391583/jobs/5848201027

mcmire: > The secret is carried through despite the top-level job not being in the environment. Maybe it only checks the environment in the job where the secret is directly read 🤔

Hmm. Maybe so! Not obvious, but if it's necessary then alright.

PR #149, opened 2022-11-18
Reorganize GitHub Action workflows

Gudahtt:
The workflows have been reorganized to use one "main" workflow for any push to the main branch, with additional jobs being triggered from there using `workflow_call`. This mirrors the change made recently in the controllers repository as part of the monorepo branch.

This makes more sense semantically every since we moved the publish step to be triggered by a push to `main`. Since then, it seend wrong to call that workflow "build-test" because it also did publishing. Now the "build-test" workflow is distinct, and it's the "main" workflow that triggers publishing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,5 +52,3 @@ jobs:
     uses: ./.github/workflows/publish-release.yml
     secrets:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-
-
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - check-workflows
-      - lint-build-test
+      - build-lint-test
     steps:
       - run: echo "Great success!"
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,56 @@
+name: Main
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  check-workflows:
+    name: Check workflows
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Download actionlint
+        id: download-actionlint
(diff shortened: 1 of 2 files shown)


legobeat (approved):

Gudahtt:
Oh right. We're using `pull_request` here, not `pull_request_target`, so this might not be triggering CI properly.

mcmire:
Should this be `pull_request_target`? It seems like there is no real difference compared with `pull_request` based on what we want to do here.

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Download actionlint
+        id: download-actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/7fdc9630cc360ea1a469eed64ac6d78caeda1234/scripts/download-actionlint.bash) 1.6.22
+        shell: bash
+      - name: Check workflow files
+        run: ${{ steps.download-actionlint.outputs.executable }} -color
+        shell: bash
+
+  build-lint-test:
+    name: Build, lint, and test
```
mcmire, line 23: What do you think about running this job after running `check-workflows`, in case that fails, so we can see the failure sooner? It shouldn't take that long to run that.

Gudahtt: Wouldn't that result in us seeing errors more slowly? Not sure I follow

Gudahtt:
Ah, I see why it wasn't running. There was a mistake in `main.yml`.

I was getting mixed up, it's `pull_request_target` that doesn't run on PRs. `pull_request` should work.

.github/workflows/main.yml
mcmire: I mean, I guess it depends on which errors we are looking for. `check-workflows` looks for specific things that may not be visible when running the workflows (if I am understanding the value of this tool correctly). So `build-lint-test` may fail, but in a different way than `check-workflows`. I guess there's no harm in that. Okay, no problem then.

mcmire (approved):
Looks good!

.github/workflows/main.yml
Gudahtt: What I mean is that if neither depends on the other, they both run without delay and produce output. But if we wait for actionlint to run before the other checks, then the other checks will be delayed or may never run at all in the case of a workflow error.

PR #151, opened 2022-11-18
Ensure required status check always runs

Gudahtt:
Apparently GitHub allows merging PRs with required status checks that have been skipped. This makes our CI checks ineffective since we've split them into separate jobs, because one job failing results in subsequent jobs being skipped (including "All jobs passed")

This has been corrected by ensuring the "All jobs passed" job always runs. This job now checks to see whether the "All jobs completed" job ran, which indicates whether anything has failed.

This was based upon the solution described in this blog post: https://brunoscheufler.com/blog/2022-04-09-the-required-github-status-check-that-wasnt


Gudahtt:
This PR demonstrates that this works: https://github.com/MetaMask/metamask-module-template/pull/150

.github/workflows/main.yml
```diff
@@ -23,14 +23,33 @@ jobs:
...
     steps:
-      - run: echo "Great success!"
+      - name: Check that all jobs have passed
+        run: |
+          passed="${{ needs.all-jobs-completed.outputs.PASSED }}"
+          if [[ $passed == "true" ]]; then
+            exit 0
+          else
+            exit 1
+          fi
```
legobeat, line 52: not a blocker but perhaps redundant with explicit `exit 0`?
```suggestion
          if [[ $passed != "true" ]]; then
            exit 1
          fi
```
or even
```suggestion
          [[ $passed != "true" ]] && exit 1
```

legobeat (dismissed):

.github/workflows/main.yml
Gudahtt: Good point!

legobeat (approved):

mcmire:
For posterity, the URL above seems to time out now, but is archived here: https://web.archive.org/web/20230402203815/https://brunoscheufler.com/blog/2022-04-09-the-required-github-status-check-that-wasnt

PR #156, opened 2022-12-09
Ignore `.yarn` folder in ESLint

Mrtenz:
The `.yarn` folder contains auto-generated files, and should not be linted by ESLint.

I've added the folder to the `ignorePatterns` to prevent this.


Gudahtt (approved):
LGTM!

PR #160, opened 2022-12-13
Lint dependencies in CI

Mrtenz:
In the extension, and some other repositories, we use `depcheck` to lint dependencies. This was never added to the module template however. It can be useful to remove unnecessary dependencies, so I've changed it to run in CI.


package.json
```diff
@@ -19,6 +19,7 @@
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
+    "lint:dependencies": "depcheck --config=.depcheckrc.json",
```
mcmire, line 22: Nit: It doesn't appear that this `--config` option is required.

mcmire (commented):
Makes sense! Is there a reason why this couldn't be incorporated into `yarn lint` though? Sorry if you tried to explain in the commit message and I didn't understand.

package.json
Mrtenz: Ah. I was having some issues when trying to use `.depcheckrc` (without extension) at first, so I tried it with the `--config` option. But you're right that it's not required now.

Mrtenz:
> Makes sense! Is there a reason why this couldn't be incorporated into `yarn lint` though?

Not really. 😅 That makes more sense. I've removed the separate workflow and added depcheck to `yarn lint` (and `yarn lint:fix`, even though it doesn't fix anything).

Gudahtt (approved):
LGTM!

rickycodes (dismissed):

mcmire (approved):

Gudahtt (approved):

PR #161, opened 2022-12-14
Prefer Babel for Jest coverage

FrederikBolding:
`babel` is the default value for `coverageProvider` and since we've seen some weird behaviour with `v8` I propose we use the default instead.


Gudahtt (approved):
LGTM!

PR #171, opened 2023-02-07
Use PAT for publishing docs

### LINKED ISSUE #184: Add branch protections for gh-pages branch
Our GitHub Actions workflows should be compatible with `gh-pages` having branch protections. Specifically, we would like to continue automatically publishing to `gh-pages` as we currently do, but require an approved PR for any manual changes to `gh-pages`.

Gudahtt:
The `publish-docs` workflow has been updated to use a personal access token for publishing docs instead of using the `GITHUB_TOKEN`. This allows for more control over doc publishing permissions.

Fixes #184

## Examples

This has been implemented in the `utils` library: https://github.com/MetaMask/utils/pull/96

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,3 +73,4 @@ jobs:
     uses: ./.github/workflows/publish-release.yml
     secrets:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      PUBLISH_DOCS_TOKEN: ${{ secrets.PUBLISH_DOCS_TOKEN }}
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -6,6 +6,9 @@ on:
       destination_dir:
         required: true
         type: string
+    secrets:
+      PUBLISH_DOCS_TOKEN:
+        required: true
 
 jobs:
   publish-docs-to-gh-pages:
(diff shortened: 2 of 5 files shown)


Gudahtt:
I was investigating whether we could[ publish using a custom GitHub actions workflow](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) rather than use a branch, and rely upon the environment required reviewers setting instead of PR review.

But unfortunately that API doesn't seem well suited for publishing to a subdirectory. The only option is to overwrite the entire site. We could download the current site and merge changes into it, except that I can't see a way to download the current site either. So long as we're publishing to subdirectories, we're stuck with the `gh-pages` branch.

mcmire (approved):
Okay, makes sense to me.

Gudahtt:
I have just updated this PR to forward secrets to the publishing workflow, which wasn't being done before. I expect that this would have failed beforehand, because the secret would have been missing.

I haven't tested this with a fork yet, I'll do that before bringing this back out of draft again. I'll add an example as well, I noticed that step was missed too.

Gudahtt:
Marking this as ready for review because it has now been tested in the utils repo.

Mrtenz (approved):

mcmire (approved):
Nice!

PR #173, opened 2023-02-10
Update ESLint test config to include Node.js rules

Gudahtt:
The ESLint configuration for test modules now includes Node.js lint rules. Our jest tests are always run with Node.js.

This was done recently in https://github.com/MetaMask/eth-json-rpc-provider/pull/6 to allow the use of Node.js globals in tests.


mcmire (approved):
Makes sense!

Mrtenz (approved):

PR #174, opened 2023-02-15
Add changelog validation to linting pipeline

mcmire:
The changelog needs to be valid in order to pass CI. To catch CI failures early, run it as a part of the `lint` script.

~Also rename `lint:misc` to `lint:prettier`, as the `--check` and `--write` options only make sense for the `prettier` executable.~


package.json
```diff
@@ -15,11 +15,12 @@
...
     "build:clean": "rimraf dist && yarn build",
     "build:docs": "typedoc",
-    "lint": "yarn lint:eslint && yarn lint:misc --check && yarn lint:dependencies",
+    "lint": "yarn lint:eslint && yarn lint:prettier --check && yarn lint:dependencies && yarn lint:changelog",
+    "lint:changelog": "auto-changelog validate",
     "lint:dependencies": "depcheck",
     "lint:eslint": "eslint . --cache --ext js,ts",
-    "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write && yarn lint:dependencies",
-    "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
+    "lint:fix": "yarn lint:eslint --fix && yarn lint:prettier --write && yarn lint:dependencies",
```
Gudahtt, line 22: Should this include `lint:changelog` as well? There is no auto-fixer, but the same is true of `lint:dependencies`.

package.json
```diff
@@ -15,11 +15,12 @@
...
+    "lint:prettier": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
```
Mrtenz, line 23: Not sure if it makes sense, but maybe we should name these scripts agnostic to the tool it's using, e.g., `lint:format`.

package.json
mcmire: I guess it doesn't hurt. Added in [f3bd4e8](https://github.com/MetaMask/metamask-module-template/pull/174/commits/f3bd4e8578f1bfade0f14c896b521d9ab5bc3594).

Gudahtt (approved):
LGTM!

package.json
Gudahtt: `lint:lint` isn't _so_ bad. Maybe `lint:code` could work instead though. Of course Prettier also affects code, but maybe it would be clear enough that code linting and formatting are separate but overlapping tasks.

legobeat: how about `lint:src` or `lint:source`?

mcmire: Hmm, I'm not sure. I'm going to change `lint:prettier` back to `lint:misc` (since it doesn't belong here anyway) and think more about this.

Gudahtt (approved):

Mrtenz (approved):

PR #175, opened 2023-02-15
Add optional Git hook to run lint before pushing

mcmire:
Say you've completed work on your branch and now you want to create a pull request, but unbeknownst to you, some code doesn't pass the lint step. It's nice to address those lint violations now, otherwise you will end up waiting until CI runs and puts an X on your PR.

Some of us on the Shared Libraries team have been testing using a Git hook to run the lint step in the `core` repo, and it has been working well so far, so this commit brings the same idea over. The Git hook runs before each push, allowing commit creation to remain speedy, and it is a completely optional part of setting up your development environment. You can install it by running `yarn sync-git-hooks`.


package.json
```diff
@@ -21,9 +21,13 @@
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write && yarn lint:dependencies",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
     "prepack": "./scripts/prepack.sh",
+    "sync-git-hooks": "simple-git-hooks",
     "test": "jest && jest-it-up",
     "test:watch": "jest --watch"
   },
+  "simple-git-hooks": {
+    "pre-push": "yarn lint"
```
Mrtenz, line 29: I would prefer this to be `pre-commit`. If you separate the commands a little bit, it could run `eslint --fix` and `prettier --write` as well, using `lint-staged`. See [snaps-monorepo](https://github.com/MetaMask/snaps-monorepo) for example.

legobeat: If changing to `pre-commit`, maybe we can add a note in `README.md` on the existence of `git commit -n` to make onboarding smoother?

mcmire: I definitely agree with using some kind of mechanism to lint fewer files. That would go a long way in improving the developer experience, I believe, and would possibly remove the concerns I have about using a pre-commit hook. `lint-staged` requires that you use a `.prettierignore` file which I'd like to avoid. I figured out some other way of doing this for another project, however, so I'll see if I can find that.

AbuSabreen: H

Jamstar00 (approved):

mcmire:
I am going to close this as it's very stale now, and this isn't a high priority.

Jamstar00 (commented):
Get hook

PR #176, opened 2023-02-24
Update GitHub Actions docs action to remove dependency on Node.js 12

Mrtenz:
I've updated `peaceiris/actions-gh-pages` to a more recent version which does not use Node.js 12 anymore for running the action. Node.js 12 has been end-of-life for some time now.


Gudahtt (approved):
LGTM!

mcmire (approved):

PR #180, opened 2023-03-08
Perform a clean build before publishing

mcmire:
This change was suggested in another PR. It helps to ensure that no unwanted files creep into new releases.

## Examples

This change is being used (or is in the process of being used) at least in this repos:

* https://github.com/MetaMask/eth-trezor-keyring/pull/161/files#diff-6378efed9843669c99725fe6165ce7eb7e9d8bea297835bd469f0357e0f94444
* https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/174/files#diff-6378efed9843669c99725fe6165ce7eb7e9d8bea297835bd469f0357e0f94444

### DIFF
--- a/scripts/prepack.sh
+++ b/scripts/prepack.sh
@@ -9,4 +9,4 @@ if [[ -n $SKIP_PREPACK ]]; then
   exit 0
 fi
 
-yarn build
+yarn build:clean


Mrtenz:
Publish runs in a separate workflow. What kind of unwanted files are in that workflow?

Gudahtt (approved):
LGTM!

Mrtenz (approved):
> This prepack script is not actually used in CI; it's skipped there. It's run only for local publishing.

I wasn't aware we did local publishing for repos based on `metamask-module-template`. Then this change makes sense to me!

Gudahtt:
The only time lately I've done local publishing for repos generated with this template is for the first published version. The package-scoped NPM token we use for automated publishing can only be generated after the package exists on npm already.

We'd likely still resort to local publishing for backports as well, as we don't have an automated publishing flow for that. Haven't needed to do that since this template was written though.

mcmire:
Good points! I will amend the commit message to mention that this is useful only for a local publishing workflow.

PR #181, opened 2023-03-09
Update TS configuration

gantunesr:


mcmire (approved):
This looks good to me! If you can update the PR description for why this is needed then that would be 🆒

PR #182, opened 2023-03-21
Generate source maps for `.d.ts` files

mcmire:
It is a frustrating experience to want to see the source code of a module that's being imported in one's editor only to be presented with the `.d.ts` file when the module is Cmd-clicked (or "Go to Definition" is invoked). To fix this, TypeScript needs to be configured to generate source maps alongside `.d.ts` files.

## Examples

I can enable this in the `core` repo first to see how this would work in practice.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "declaration": true,
+    "declarationMap": true,
     "inlineSources": true,
     "noEmit": false,
     "outDir": "dist",


Mrtenz (approved):
Seems like a good idea regardless to add this. 👍

mcmire:
PR for trying this out in `core`: https://github.com/MetaMask/core/pull/1161

mcmire:
Update: It is now possible to jump from a symbol in VSCode to the compiled `.js` file that holds the implementation for the symbol rather than the type definition file.

However, VSCode has a **separate command** to achieve this than the "Go to Definition" command that it usually uses when you Cmd-/Ctrl-Click on a symbol. Rather, it's called "Go to Source Definition" instead. You can either call this by opening up the Command Palette and choosing this, or you can get VSCode to use this command by default by going into Settings and checking these two boxes:

Screenshot 2023-06-06 at 9 26 07 PM

Once you do that then everything should just work.

Gudahtt (approved):
LGTM!

PR #186, opened 2023-04-19
Changed node version in packagejson to match nvmrc

BelfordZ:
The nvmrc file specifies v16, the package.json#engines says greater than 14. This PR makes them both 16.

as an aside, using un-specific versions in nvmrc is discouraged. We should be specifying the exact version of node to be used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   },
   "packageManager": "yarn@3.2.1",
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
   },
   "publishConfig": {
     "access": "public",


Gudahtt:
Maybe better to title this as "Increase minimum supported Node.js version to v16".

The nvmrc and the minimum version are _intentionally_ out of sync in some projects, where a newer Node.js version provides a better development experience despite there being an intent to support older versions. But I agree it's simpler to keep them in sync.

Gudahtt (changes requested):
Looks good except that the README also refers to v14. Could you update that as well?

Gudahtt:
>using un-specific versions in nvmrc is discouraged. We should be specifying the exact version of node to be used.

Could you elaborate on why this is discouraged?

Gudahtt:
I have created a new PR to handle this bump plus the v19 => v20 bump: https://github.com/MetaMask/metamask-module-template/pull/187

BelfordZ:
the purpose of nvmrc is to make sure we are all running the same node version. By specifying `v${NUMBER}`, its unclear what the actual node version will be. Further, the node version can change without the nvmrc file having changed.

Also if you use specific versions, you dont need to go figure out what v16 resolves to. 

nodenv doesn't support using unspecific versions like this, one of the reasons why it is a superior tool for this problem.

PR #191, opened 2023-04-24
Add constraints to enforce common `package.json` fields

Mrtenz:
## Description

This pull request adds constraints to enforce common `package.json` fields. This will make sure that:

1. The package must have a `name`.
2. The package must have a `description`.
   - The description cannot end with a period. This is how it's set in `MetaMask/core`, but I'm open to change this.
3. The package must have a `homepage`.
   - The homepage must be set to `https://github.com/MetaMask/package-name#readme`.
4. The package must have a `bugs.url`.
   - The bugs URL must be set to `https://github.com/MetaMask/package-name/issues`.
5. The package must have a `repository`.
   - The repository type must be set to Git.
   - The repository URL must be set to `https://github.com/MetaMask/package-name.git`.
6. The package must have a `license`.
7. The package must have a `main`.
   - The main entry point must be set to `./dist/index.js`.
8. The package must have `types`.
   - The types entry point must be set to `./dist/index.d.ts`.
9. The package must specify `files`.
   - The files must be set to `['dist/']`.
10. The package must specify a minimum Node.js version of `>=16.0.0`.
11. The package must have a `publishConfig`.
    - The `access` field must be set to `public`.
    - The `registry` field must be set to `https://registry.npmjs.org/`.
12. Dependencies may not appear in both `dependencies` and `devDependencies`.

This is largely based on the constraints used in the core repository (thanks @mcmire!), but adapted for a single package, rather than a monorepo.

This will prevent issues such as:

- Deployment failing because `repository` is a `string`, rather than an `object` containing the type and URL.
- The package is deployed as a private package, because `publishConfig` is not set.
- The package does not specify important fields, such as `main`, `types`, or `files`.

## Changes

1. Add constraints file


FrederikBolding (dismissed):

constraints.pro
```diff
@@ -0,0 +1,105 @@
...
+  last(Parts, WorkspaceBasename).
+
+% True if and only if WorkspacePackageName can unify with the name of the
+% package which the workspace represents (which comes from the directory where
+% the package is located). Assumes that the package is not in a sub-workspace
+% and is not private.
+workspace_package_name(WorkspaceCwd, WorkspacePackageName) :-
+  workspace_basename(WorkspaceCwd, WorkspaceBasename),
+  atom_concat('@metamask/', WorkspaceBasename, WorkspacePackageName).
+
```
Gudahtt, line 18: It seems that these utility predicates were only used for monorepo-related things (e.g. ensuring the package directory matches the package name). They don't seem to be used here.

```suggestion
```

Gudahtt (commented):
Looks great! I found what appears to be dead code in the constraints file; I can approve once that is resolved.

constraints.pro
Mrtenz: You're right. I tried removing them initially, but something was depending on them, but I think I removed that later. Fixed it now!

FrederikBolding (approved):

PR #192, opened 2023-04-26
Run `yarn lint:constraints` before `yarn lint:misc`

Mrtenz:
## Description

This pull request changes the order of two commands in the `lint` script defined in `package.json`. The `lint:constraints`
command is moved before `lint:misc` command to avoid any conflicts caused by Yarn constraints in the latter command.

This is especially important when using `yarn lint:fix`. If `yarn constraints` makes any modifications to `package.json`, these would not be formatted by Prettier before.

## Changes

1. Change the order of the `lint` commands in the `package.json` file.
   - Now `lint:constraints` runs before `lint:misc`.


Gudahtt (approved):
LGTM!

PR #196, opened 2023-05-15
Add Slack webhook config

rickycodes:
Adding this to the template


.github/workflows/publish-release.yml
```diff
@@ -5,6 +5,8 @@ on:
     secrets:
       NPM_TOKEN:
         required: true
+      SLACK_WEBHOOK_URL:
```
Gudahtt, line 8: This needs to be passed in at the point where this workflow is invoked

Gudahtt (approved):
LGTM!

mcmire (approved):
Looks good!

PR #197, opened 2023-05-18
devDeps: use `rm` command instead of rimraf package

legobeat:
The POSIX `rm` command can be assumed I think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   ],
   "scripts": {
     "build": "tsc --project tsconfig.build.json",
-    "build:clean": "rm -rf dist && yarn build",
+    "build:clean": "rm -r dist && yarn build",
     "build:docs": "typedoc",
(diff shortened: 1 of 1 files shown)


Gudahtt:
The intention was to support Windows for any commands a contributor might use. We only assume a POSIX/Linux environment for commands intended just for CI.

mcmire:
@legobeat If we're not using `-f` then what do you think about checking that the directory exists first before deleting it? I'm not sure that `yarn build:clean` should fail if nothing has been built yet.

Gudahtt:
It seems that [the default shell used by npm on Windows](https://docs.npmjs.com/cli/v6/commands/npm-run-script) is `cmd.exe`, so this wouldn't be enough to make this work out of the box.

Seems better to leave this as-is, especially given that `rimraf` is a requirement of other dependencies anyway.

PR #201, opened 2023-06-21
Remove the `-x` flag from Bash scripts

Gudahtt:
The `-x` flag is typically used for debugging. When used in production, it can clutter the console output and make it difficult to tell what's happening.

The flag has been removed from the two small Bash scripts in this template.

## Examples

This was done recently in this PR: https://github.com/MetaMask/eth-trezor-keyring/pull/182

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
-set -x
 set -e
 set -u
 set -o pipefail
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -4,6 +4,10 @@ set -e
 set -u
 set -o pipefail
 
+if [[ ${RUNNER_DEBUG:-0} == 1 ]]; then
+  set -x
+fi
+
 KEY="${1}"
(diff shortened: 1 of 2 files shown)


mcmire (commented):
I agree that usually this output isn't necessary, but it could be necessary when debugging actions. What are your thoughts on having some way to enable `-x` but only when the "Enable debugging" option is turned on? It looks like when that happens, `runner.debug` in the context will be set: https://stackoverflow.com/questions/75434105/how-to-determine-if-the-github-action-is-running-with-debug-logging-enabled

Gudahtt:
Sure, that might make sense for `get.sh` at least. `prepack.sh` is local-only, so it would always be unset there.

Gudahtt:
I have updated the script to set `-x` if RUNNER_DEBUG is set (see https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

Mrtenz (approved):

mcmire (approved):
LGTM!

PR #203, opened 2023-06-22
Build package as both CJS and ESM

Mrtenz:
Instead of transpiling the source code to just CJS, we now transpile it to ESM as well. This is beneficial for projects using bundlers that are capable of tree-shaking (such as MetaMask Snaps), because it's much easier to determine whether the code has side effects or not.

For the transpilation, we now use [SWC](https://swc.rs/) rather than TSC. This avoids having to do type checking twice, which is quite slow. TypeScript is still used for generating the declaration files, however, so the code is still being type-checked.

This is based on a change in the Snaps repository (MetaMask/snaps#1519), which has been tested in the extension, so this shouldn't cause any problems there.


.github/workflows/build-lint-test.yml
```diff
@@ -33,7 +33,32 @@ jobs:
...
+          fi
+
+  build-types:
+    name: Build types
+    runs-on: ubuntu-latest
+    needs:
+      - prepare
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x, 20.x]
```
FrederikBolding, line 52: We probably don't need to run build on all these Node versions. But can be fixed in another PR.

FrederikBolding (approved):

.github/workflows/build-lint-test.yml
Mrtenz: Agreed, but out of scope for this PR.

.swcrc.build.json
```diff
@@ -0,0 +1,20 @@
...
+  "$schema": "https://json.schemastore.org/swcrc",
+  "jsc": {
+    "parser": {
+      "syntax": "typescript"
+    },
+    "target": "es2020"
+  },
+  "sourceMaps": true,
+  "exclude": [
+    ".*__fixtures__.*",
```
mcmire, line 13: `tsconfig.build.json` only excludes test files; it doesn't exclude files related to tests. Should we sync `tsconfig.build.json` with this list? Or does it matter?

package.json
```diff
@@ -10,15 +10,29 @@
     "type": "git",
     "url": "https://github.com/MetaMask/metamask-module-template.git"
   },
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "exports": {
```
mcmire, line 13: Will this change how libraries can be exported going forward? Currently we are allowed to do `import { X } from "@metamask/y/dist/z"`.

constraints.pro
```diff
@@ -65,15 +65,15 @@ gen_enforced_field(WorkspaceCwd, 'repository.url', 'https://github.com/MetaMask/
 % The license for the package must be specified.
 gen_enforced_field(WorkspaceCwd, 'license').
 
-% The entrypoint for the package must be `./dist/index.js`.
-gen_enforced_field(WorkspaceCwd, 'main', './dist/index.js').
+% The entrypoint for the package must be `./dist/cjs/index.js`.
+gen_enforced_field(WorkspaceCwd, 'main', './dist/cjs/index.js').
```
mcmire, line 69: Should we add more constraints to check:

- `module`
- `exports`

mcmire: (I'm also noticing that we don't have constraints to check `scripts` like we do in `core`, but that's out of scope)

tsconfig.build.json
```diff
@@ -3,9 +3,10 @@
   "compilerOptions": {
     "declaration": true,
     "declarationMap": true,
```
mcmire, line 5: If we're only emitting type definitions, would we still be emitting source maps for those type definitions?

mcmire (commented):
Very cool! Just had a few comments/suggestions.

.swcrc.build.json
Mrtenz: Good catch, we should sync these. There's no point in generating types for something that we don't have JS files for. 😅

package.json
Mrtenz: This is still allowed, but consumers have to pick whether they want the CJS or ESM version, i.e., `import { X } from '@metamask/y/dist/cjs/z'` or `'@metamask/y/dist/esm/z'`.

tsconfig.build.json
Mrtenz: TypeScript still emits a source map. I'm not sure how to test if it's correct, but it seems alright?

```json
{"version":3,"file":"index.d.ts","sourceRoot":"","sources":["../../src/index.ts"],"names":[],"mappings":"AAAA;;;;;GAKG;AACH,MAAM,CAAC,OAAO,UAAU,OAAO,CAAC,IAAI,EAAE,MAAM,GAAG,MAAM,CAEpD"}
```

Gudahtt: Hmm, I found this unexpected as well, as I don't know of a time I've needed to map a type to its source code. But I guess that makes sense.

package.json
Gudahtt: @mcmire is correct, this does prevent referencing files inside the package. See https://nodejs.org/api/packages.html#package-entry-points

>The ["exports"](https://nodejs.org/api/packages.html#exports) provides a modern alternative to ["main"](https://nodejs.org/api/packages.html#main) allowing multiple entry points to be defined, conditional entry resolution support between environments, and **preventing any other entry points besides those defined in ["exports"](https://nodejs.org/api/packages.html#exports)**. This encapsulation allows module authors to clearly define the public interface for their package.

Gudahtt: Though I am OK with this being in our template. We try not to rely on referencing files directly in packages, since we tend not to consider that in determining which changes are breaking.

Mrtenz: Interesting. I was testing this in the extension, and it worked there, but that may be Browserify using their own `require` implementation. 🤔

Mrtenz: Adding an export for `./*` seems to resolve this.

tsconfig.build.json
Mrtenz: More context here: #182.

mcmire: Right — you should be able to test this by clicking on a symbol in an import and issuing the "Go to Source Definition" command and you should be directed to the JS file rather than the type definition file (follow the directions here: https://github.com/MetaMask/metamask-module-template/pull/182#issuecomment-1579817679).

package.json
mcmire: I'm okay with this being in our template as well, but maybe we reserve this for another PR just for visibility's sake.

mcmire (dismissed):
Looks good!

Mrtenz:
After some more testing, I found that Browserify does not support the `exports` field (browserify/resolve#222). This means that we can't use any scoped imports, as they will not work in the extension. For example, if `@metamask/foo` contains the following import:

```ts
import { bar } from '@metamask/bar/baz'; // dist/cjs/baz/index.js
```

This will work fine in Node.js, Webpack, etc., but in Browserify this results in an error:

> Could not resolve "@metamask/bar/baz"

I have an idea on how this can be solved, but will put it into draft until this has been tested.

Gudahtt:
Could you elaborate on the connection between that problem, and the contents of this PR? I see that this PR adds the `exports` field, but it's less clear to me how this impacts browserify (i.e. does it matter that browserify can still only use the `main`/`module` entrypoint?)

Mrtenz:
> Could you elaborate on the connection between that problem, and the contents of this PR? I see that this PR adds the `exports` field, but it's less clear to me how this impacts browserify (i.e. does it matter that browserify can still only use the `main`/`module` entrypoint?)

Sure. The problem is that packages that don't use Browserify could import in ways that are not supported by Browserify. For example, if `@metamask/utils` imports from `@metamask/key-tree/some/path`, and `@metamask/utils` gets imported into the extension, Browserify would fail because `node_modules/@metamask/key-tree/some/path` does not exist. It doesn't look at the `exports` field to try and resolve the proper path, i.e., `node_modules/@metamask/key-tree/dist/cjs/some/path`.

We solved this in the Snaps repo by explicitly exporting `./dist/cjs/*` and `./dist/esm/*`, rather than `./*`. This means that `@metamask/utils` can import `@metamask/key-tree/dist/cjs/some/path`, because it's specified in the `exports` field, and Browserify can import it, because it's an actual path on the disk: `node_modules/@metamask/key-tree/dist/cjs/some/path` exists.

Ideally we avoid importing from `package/dist` in the first place, but since we're doing that, I think this is the best solution as long as we're still using Browserify.

I hope that makes sense!

Gudahtt:
Thanks! Makes sense.

Perhaps in the meantime we could use the `main` and `module` fields to support multiple build types? Reserving the nicer export entry points for later, when that problem is resolved.

Mrtenz:
> Thanks! Makes sense.
> 
> Perhaps in the meantime we could use the `main` and `module` fields to support multiple build types? Reserving the nicer export entry points for later, when that problem is resolved.

I hadn't even considered that we don't need `exports` at all in that case. 😅

FrederikBolding (dismissed):

mcmire (dismissed):
LGTM!

.swcrc.build.json
```diff
@@ -0,0 +1,20 @@
+{
+  // Note that this file is intentionally called `.swcrc.build.json`, so that if
```
Gudahtt, line 2: Nit: Comments aren't generally allowed in JSON. I see that SWC does allow them though.

Perhaps we could change the filename to stop editors/linters from complaining about this? The default SWC config seems to be called `.swcrc`, so perhaps we could call this `.build.swcrc`

Gudahtt (approved):
LGTM!

Gudahtt:
@SocketSecurity ignore-all

The remaining alerts are regarding packages used by SWC with access that seems reasonable for them to have, and new author/unmaintained warnings (none of which seem suspicious).

.swcrc.build.json
Gudahtt: Sure, good point. Maybe removing the comments is fine then.

Gudahtt: Or we could use `.json5` as the suffix

.swcrc.build.json
```diff
@@ -0,0 +1,18 @@
+{
```
Mrtenz, line 1: Just leaving a comment here for future reference.

When we change the package to use SWC for other parts of the stack, such as testing, we need to keep in mind to use a different `.swcrc` file. This one is explicitly ignoring test files and such, so will not work with Jest.

mcmire (approved):

Gudahtt (approved):

PR #204, opened 2023-06-22
Add `sideEffects: false` to `package.json`

Mrtenz:
By setting `sideEffects: false`, we can mark the package as side-effect-free:

https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

This will improve the effectiveness of tree-shaking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ Follow these instructions when using this template.
 - Update the repository URL, referenced in the following places:
   - `repository` field of `package.json`
   - The links in the API section of the README
+- If your project has side effects, update the `sideEffects` field of
+  `package.json` to `true`, or an array of file globs that match the files that
+  have side effects.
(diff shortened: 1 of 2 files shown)


FrederikBolding (approved):

wachunei:
I'm concerned about the cases a package is created using this template that actually has side effects and the import of the side effect gets removed, maybe we should have a explicit `README` section to remember to change this option.

Mrtenz:
> I'm concerned about the cases a package is created using this template that actually has side effects and the import of the side effect gets removed, maybe we should have a explicit `README` section to remember to change this option.

Good point. I've added a mention to the README.

FrederikBolding (approved):

mcmire (approved):
Makes sense!

PR #214, opened 2023-07-24
Fix broken Dependabot link

Mrtenz:
The old link simply redirected to `support.github.com`. I've updated it to the new GitHub Docs link.


Gudahtt (approved):
LGTM!

PR #215, opened 2023-07-26
ci: Run compatibility test ignoring lockfile

legobeat:
This adds an additional run of tests against a state where `yarn.lock` is removed prior to the installation. This is intended to catch e.g. regressions and mislabeled breakage in updated transitive dependencies.

This will map more closely to what users get when installing the released package from registry.

Left out the intermediate versions of node as an optimization. The normal cache is still used since significant overlap is expected.

### Potential future improvements

- Run test against built package under another package manager (yarn v1 and/or npm)
  - Ensure there aren't implicit dependencies on yarn berry
  - Would allow throwing on breakages in `engines` without adding a plugin

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -124,7 +124,7 @@ jobs:
           fi
 
   compatibility-test:
-    name: Compatibility Test
+    name: Compatibility test
     runs-on: ubuntu-latest
     needs:
       - prepare


.github/workflows/build-lint-test.yml
```diff
@@ -122,3 +122,28 @@ jobs:
             echo "Working tree dirty at end of job"
             exit 1
           fi
+  compat_test:
```
Mrtenz, line 125:
```suggestion

  compatibility-test:
```

.github/workflows/build-lint-test.yml
```diff
@@ -122,3 +122,28 @@ jobs:
...
+    name: Compatibility Test
```
Mrtenz, line 126:
```suggestion
    name: Compatibility test
```

Mrtenz (changes requested):
Not sure I fully understand the need for this, but either way here's some nits. 😅

Mrtenz (approved):
Makes sense!

mcmire (approved):
Looks good!

PR #216, opened 2023-07-27
Update node.engines field in package.json

legobeat:
This should more correctly reflect actual intended support in most cases:

- Odd-numbered versions are effectively ephemeral and supporting anything but the most recent doesn't seem meaningful
- Signaling recent minors for maintenance/LTS in means less overhead in supporting the older runtimes.


constraints.pro
```diff
@@ -90,8 +90,8 @@ gen_enforced_dependency(WorkspaceCwd, DependencyIdent, null, DependencyType) :-
   workspace_has_dependency(WorkspaceCwd, DependencyIdent, DependencyRange, DependencyType),
   DependencyType == 'devDependencies'.
 
-% The package must specify a minimum Node version of 16.
-gen_enforced_field(WorkspaceCwd, 'engines.node', '>=16.0.0').
+% The package must specify the expected node version contraint.
```
mcmire, line 93: Typo:

``` suggestion
% The package must specify the expected node version constraint.
```

package.json
```diff
@@ -82,7 +82,7 @@
   },
   "packageManager": "yarn@3.2.1",
   "engines": {
-    "node": ">=16.0.0"
+    "node": "^16.20 || ^18.16 || >=20"
```
mcmire, line 85: Small thing, but is it necessary to be so strict about the version of 16.x and 18.x? Does it make sense to use `^16 || ^18 || >=20` now, and then carry that trend forward, so when 22 comes out we shift it to `^18 || ^20 || >=22`? I suppose we could simplify this later, but if we can do it now then it seems better to me.

mcmire: We had a discussion about this over a call and also on Slack and have resolved this. This strategy makes sense provided that we now use this range as a template for other repos going forward (except where necessary to diverge).

mcmire (approved):
This looks good to me. Let's do it!

Gudahtt (approved):
LGTM!

PR #219, opened 2023-09-28
Enable Prettier for changelog validation

Mrtenz:
This enables Prettier for changelog validation.

Blocked by:

- MetaMask/action-create-release-pr#116


mcmire:
This should be unblocked now.

mcmire (dismissed):
Looks good!

mcmire:
Hmm. It doesn't look like the CHANGELOG needs that extra line for some reason.

Inikowu (approved):

Inikowu:
[Inikowu-Isaac.zip](https://github.com/MetaMask/metamask-module-template/files/13473961/Inikowu-Isaac.zip)
package.json

mcmire:
It looks like when we validate the changelogs in CI, we manually run `yarn auto-changelog`. Perhaps we should not do this anymore? I see that we are doing this because we may need to pass `--rc` depending on whether the current branch is a release branch or not, but we could extract the existing `lint:changelog` package script to a shell script and then add the same `if` statement that exists in the CI config. That way it gets run consistently locally and in CI.

Mrtenz:
> It looks like when we validate the changelogs in CI, we manually run `yarn auto-changelog`. Perhaps we should not do this anymore? I see that we are doing this because we may need to pass `--rc` depending on whether the current branch is a release branch or not, but we could extract the existing `lint:changelog` package script to a shell script and then add the same `if` statement that exists in the CI config. That way it gets run consistently locally and in CI.

I just swapped it out for `yarn lint:changelog`, which seems to work fine, and seems easier than adding a script.

mcmire:
Ah I think the PR I merged might have created merge conflicts, sorry.

mcmire (approved):
Looks good!

bowensanders (approved):
LGTM

PR #220, opened 2023-09-30
Implement JavaScript-based constraints

Mrtenz:
I've replaced the Prolog-based constraints (deprecated in Yarn v4) with JavaScript-based constraints. This has the benefit that the constraints file is much more readable, especially for those who are not well-versed in Prolog. As an added benefit, the error messages produced by Yarn are more readable too.

The new constraints file implements all the Prolog-based constraints, and some extras. For example, we can now verify that the `README.md` does not contain the template instructions (if the current repository is not the module template), and that the `.nvmrc` version matches with the Node.js version mentioned in the `README.md`. Since the new constraints file has access to all Node.js APIs (including `fs`), we have much more flexibility in the kinds of constraints.

Using these constraints requires bumping Yarn to v4, and using a minimum Node.js version of 18.12.0. The Node.js version bump is a `.nvmrc` only bump, so this shouldn't be a breaking change.

### Blocked by

- #221


Mrtenz:
> Using these constraints requires bumping Yarn to v4, and using a minimum Node.js version of 18.12.0. The Node.js version bump is a .nvmrc only bump, so this shouldn't be a breaking change.

Unfortunately this causes CI to fail, so it looks like we may want to drop support for Node.js 16 (EOL) anyway.

.eslintrc.js
```diff
@@ -10,9 +10,15 @@ module.exports = {
     },
 
     {
-      files: ['*.js'],
+      files: ['*.js', '*.cjs'],
       parserOptions: {
         sourceType: 'script',
+        ecmaVersion: 2020,
```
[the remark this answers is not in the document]
Mrtenz: Not sure. I've moved the Yarn config into a separate config block in the ESLint config so we can change settings for just that file.

.yarnrc.yml
```diff
@@ -1,3 +1,7 @@
+compressionLevel: mixed
```
FrederikBolding, line 1: What's the reason for these new config options?

yarn.config.cjs
```diff
@@ -0,0 +1,273 @@
...
+      const keys = Object.keys(exportObject);
+      if (keys.includes('types') && keys[0] !== 'types') {
+        workspace.error(
+          `The "types" export must be the first export in the "exports" field for the export "${exportName}".`,
+        );
+      }
+    });
+}
+
+module.exports = defineConfig({
```
FrederikBolding, line 212: So much more readable 🎉

.yarnrc.yml
```diff
@@ -11,7 +15,5 @@ nodeLinker: node-modules
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
     spec: "https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js"
-  - path: .yarn/plugins/@yarnpkg/plugin-constraints.cjs
-    spec: "@yarnpkg/plugin-constraints"
 
-yarnPath: .yarn/releases/yarn-3.2.1.cjs
+yarnPath: .yarn/releases/yarn-4.1.1.cjs
```
mcmire, line 19: Heads up: We don't need to include the Yarn binary in the repo anymore with Yarn 4; [it defaults to using Corepack now](https://yarnpkg.com/blog/release/4.0). We can make that change in another PR, but I thought I'd point that out.

yarn.config.cjs
```diff
@@ -0,0 +1,273 @@
...
+    workspace.set('exports["."].import', './dist/index.mjs');
+
+    // The package must export a `package.json` file.
+    workspace.set('exports["./package.json"]', './package.json');
+
+    expectExports(workspace);
+
+    // The list of files included in the package must only include files
+    // generated during the build process.
+    workspace.set('files[0]', 'dist');
```
mcmire, line 266: It looks like this says that `dist` must be the first item in `files`. Does that matter? Should we say that `dist` must be present in `files` instead?

yarn.config.cjs
```diff
@@ -0,0 +1,273 @@
...
+          `The "types" export must be the first export in the "exports" field for the export "${exportName}".`,
+        );
+      }
+    });
+}
+
+module.exports = defineConfig({
+  async constraints({ Yarn }) {
+    const workspace = Yarn.workspace();
+    const workspaceName = basename(__dirname);
```
mcmire, line 215: Making this dependent on the directory name makes sense, but it also means that this constraint may act differently depending on where it's run. For instance, if a repo is renamed but the developer doesn't know it and they haven't renamed their directory yet then the constraint could suddenly fail for them.

It it possible to assert that the `name` field is present, and then derive the `workspaceName` from that?

mcmire (commented):
This is the first time I'm looking at the new constraints API up close, and unfortunately it's not as readable or understandable as I'd hoped (I should really submit a pull request to the Yarn repo). That said, wrapping each step in a well-named function is a good idea, and the fact that we can read files and do other kinds of things is a big win.

I just had some comments but overall this is looking good.

yarn.config.cjs
Mrtenz: Order of `files` doesn't matter. I've changed it to `workspace.set('files', ['dist'])` to make this clear. We could also check if the array contains `dist`, but I don't think that's possible with `workspace.set`.

yarn.config.cjs
Mrtenz: I've changed it to use `workspace.ident` instead.

.eslintrc.js
```diff
@@ -10,9 +10,15 @@ module.exports = {
...
+      },
+      settings: {
+        jsdoc: {
+          mode: 'typescript',
```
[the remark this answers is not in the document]
mcmire: Ticket added here: https://github.com/MetaMask/eslint-config/issues/349. Resolving this conversation.

yarn.config.cjs
mcmire: I think that's okay. Projects can customize this if they really need to, but I imagine that will be rare.

mcmire (approved):
Looks good!

FrederikBolding (approved):

PR #229, opened 2023-11-02
Use tsup for building

Mrtenz:
This is based on MetaMask/utils#144. Refer to that PR for the rationale behind it.


constraints.pro
```diff
@@ -69,19 +69,19 @@ gen_enforced_field(WorkspaceCwd, 'license').
...
 gen_enforced_field(WorkspaceCwd, 'exports["."].types', './dist/types/index.d.ts').
 
-% The entrypoint for the package must be `./dist/cjs/index.js`.
-gen_enforced_field(WorkspaceCwd, 'main', './dist/cjs/index.js').
-gen_enforced_field(WorkspaceCwd, 'exports["."].require', './dist/cjs/index.js').
+% The entrypoint for the package must be `./dist/index.js`.
+gen_enforced_field(WorkspaceCwd, 'main', './dist/index.js').
+gen_enforced_field(WorkspaceCwd, 'exports["."].require', './dist/index.js').
 
 % The module entrypoint for the package must be `./dist/esm/index.js`.
```
FrederikBolding, line 76:
```suggestion
% The module entrypoint for the package must be `./dist/index.mjs`.
```

FrederikBolding (approved):

Mrtenz:
@SocketSecurity ignore source-map@0.8.0-beta.0
@SocketSecurity ignore esbuild@0.18.20

tsup.config.ts
```diff
@@ -0,0 +1,40 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  // The entry to bundle.
+  entry: [
+    'src/**/*.ts',
+    '!src/**/__fixtures__/**/*',
```
mcmire, line 7: It seems that we would have to keep this in sync with TypeScript and Jest configuration. I don't have any suggestions here but it would be nice if we didn't have to do this because I can see this getting forgotten. I guess the good thing is that we tend to check the files inside of a published package before it goes out.

mcmire (approved):
Looks good!

tsup.config.ts
Mrtenz: Yeah, not much we can do about that I think. I added the most common files and folders, so in most cases we won't have to modify it.

FrederikBolding (approved):

PR #230, opened 2023-11-02
Run build and lint steps in CI on `.nvmrc` Node version

Mrtenz:
This changes the build and lint steps in CI to only run on the `.nvmrc` version. The Node version for these steps shouldn't matter. The test step is still run on all supported Node versions.

We made this change in a few other repositories already, but it was never applied to the module template.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -22,15 +22,12 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - prepare
-    strategy:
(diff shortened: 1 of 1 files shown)


FrederikBolding (approved):

MajorLift (approved):
Running the build, lint steps with the node version used for development, and testing on all supported versions makes sense. LGTM!
- related: https://github.com/MetaMask/metamask-module-template/pull/222#issuecomment-1745487946

PR #231, opened 2023-11-03
Downgrade eslint-plugin-import to ~2.26.0

Mrtenz:
Versions >=2.27.0 of `eslint-plugin-import` have a regression, causing the import order to change. The recommended version in the ESLint config is `~2.26.0` for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "depcheck": "^1.4.3",
     "eslint": "^8.44.0",
     "eslint-config-prettier": "^8.8.0",
-    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-import": "~2.26.0",
     "eslint-plugin-jest": "^27.2.2",
     "eslint-plugin-jsdoc": "^39.9.1",
     "eslint-plugin-n": "^15.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,7 +55,7 @@
     "depcheck": "^1.4.3",
     "eslint": "^8.44.0",
     "eslint-config-prettier": "^8.8.0",
-    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-import": "~2.26.0",
```
mcmire, line 58: Can we go ahead and bump to `^2.29.0`? It looks like there is a fix for something in 2.28.1, but I'm not sure if it's what you're referring to: https://github.com/import-js/eslint-plugin-import/blob/main/CHANGELOG.md#fixed

Mrtenz: I checked `2.29.0` but it does not fix the issue. There are several open issues about it:

- import-js/eslint-plugin-import#2909
- import-js/eslint-plugin-import#2683
- import-js/eslint-plugin-import#2682
- And more...

It looks like this PR will fix it: import-js/eslint-plugin-import#2721

mcmire: I see. Thanks for following up!

mcmire (approved):
Looks good!

david0xd (approved):

PR #238, opened 2024-01-18
Fix 'yarn pack'

mcmire:
The `yarn pack` command is run by the `npm-publish` action prior to publishing the package. Unfortunately, this command does not work as the `prepack` lifecycle script is overridden to run `yarn build:clean`, which was removed when this repo was converted to use `tsup` for building. `tsup` already clears the `dist` directly, so in practice a `build:clean` script is not usually necessary, but there are occasions where we want to _really_ be sure that `dist` does not exist, and publishing is one of them. Hence, this commit adds back `build:clean`, which should fix `yarn pack` and the publishing workflow.

## Testing

Running `yarn pack` should not produce any errors.

## Examples

N/A, as this is a bugfix to this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsup && yarn build:types",
-    "build:clean": "rimraf dist && yarn build",
+    "build": "tsup --clean && yarn build:types",
     "build:docs": "typedoc",
     "build:types": "tsc --project tsconfig.build.json",
     "lint": "yarn lint:eslint && yarn lint:constraints && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   ],
   "scripts": {
     "build": "tsup && yarn build:types",
+    "build:clean": "rimraf dist && yarn build",
     "build:docs": "typedoc",
     "build:types": "tsc --project tsconfig.build.json",
     "lint": "yarn lint:eslint && yarn lint:constraints && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
   ],
   "scripts": {
     "build": "tsup && yarn build:types",
+    "build:clean": "rimraf dist && yarn build",
```
Gudahtt, line 30: I don't see `rimraf` in the dependency list

Gudahtt: There is a `--clean` flag for `tsup` that we could consider using instead of rimraf

Mrtenz:
> `tsup` already clears the `dist` directly, so in practice a `build:clean` script is not usually necessary, but there are occasions where we want to _really_ be sure that `dist` does not exist, and publishing is one of them. Hence, this commit adds back `build:clean`, which should fix `yarn pack` and the publishing workflow.

Not sure I understand the reasoning here. If `tsup` guarantees to clear the dist, isn't it easier to modify the prepack script to run `yarn build`?

mcmire:
@Mrtenz I wanted to make sure that we completely remove `dist` before publishing so that we can be absolutely certain that unwanted files don't slip into releases.

I checked the `tsup` source code and it does seem that everything gets removed from `dist`: https://github.com/egoist/tsup/blob/8c26e63c92711d60c05aedd3cdc358530ba266c5/src/index.ts#L303

So yes, you're right, we can just use `yarn build` in the prepack script. I've done that here: [d04b7e7](https://github.com/MetaMask/metamask-module-template/pull/238/commits/d04b7e7b90e988482a08fa8951c6e30b957a4589)

package.json
mcmire: `--clean` seems to do what we want. I've made that change here: [d04b7e7](https://github.com/MetaMask/metamask-module-template/pull/238/commits/d04b7e7b90e988482a08fa8951c6e30b957a4589)

Mrtenz (approved):

Gudahtt (approved):
LGTM!

PR #243, opened 2024-04-25
Unvendor Yarn; favor Corepack for installation

mcmire:
According to the Yarn documentation, Yarn no longer needs to be bundled with a project, but rather, installing Yarn ought to be done using [Corepack](https://yarnpkg.com/corepack), which is shipped in modern Node versions. This commit removes the Yarn binary from the repo and updates the installation instructions in the README to match.

## Examples

(None)


mcmire (commented):
I kind of dislike the hacks we have to do here. It makes the GitHub workflow files longer and more complicated than they need to be. It's true Corepack is the recommended way to install Yarn, but it doesn't mean that the `yarnPath` method is deprecated. So I'm inclined to not do this and keep the current way. Then again, the downside to the current approach is that projects are obligated to use the same version of Yarn as in the template, which doesn't seem completely necessary either. Sigh...

mcmire:
Yeah, that's true. Okay, I'm convinced. ✅

legobeat:
@mcmire ping

HowardBraham (approved):

PR #244, opened 2024-04-25
ci: run `corepack enable` before installing dependencies

legobeat:
`corepack` is not currently supported in `@actions/setup-node` and this means installation of dependencies fail when using yarn >v1:
- https://github.com/actions/setup-node/pull/901

This hacks around it by running `@actions/setup-node` twice: once before checkout to install the node-version, so that `corepack enable` can be run, after which the normal flow starting with checkout can proceed.

Since `.nvmrc` is not available before checkout, it also implies not using `.nvmrc` to specify node-version for this workflow anymore.

An alternative to this would be forking `@actions/setup-node` with the patch from the PR linked above but sounds not worth it.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -8,14 +8,18 @@ jobs:
     name: Prepare
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version-file: '.nvmrc'
-          cache: 'yarn'
+          node-version: 'lts/*'
       - name: Install Yarn
         run: corepack enable
+      - uses: actions/checkout@v4
+      - name: Use Node.js and install dependencies
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+          cache: 'yarn'
       - name: Install Yarn dependencies
         run: yarn --immutable
 
@@ -25,11 +29,17 @@ jobs:
     needs:
       - prepare
     steps:
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+      - name: Install Yarn
+        run: corepack enable
       - uses: actions/checkout@v4
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version-file: '.nvmrc'
+          node-version: 'lts/*'
           cache: 'yarn'
       - run: yarn --immutable --immutable-cache
       - run: yarn build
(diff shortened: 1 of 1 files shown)


Mrtenz:
> this means installation of dependencies fail when using yarn >v1:

Not sure I understand. We're using Yarn 3 and 4 in several repos without problems.

.github/workflows/build-lint-test.yml
```diff
@@ -8,11 +8,17 @@ jobs:
     name: Prepare
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version-file: '.nvmrc'
+          node-version: 'lts/*'
```
Mrtenz, line 14: Should we not prefer the version in `.nvmrc`?

mcmire: According to the PR description, we are now running this first before checking out, so we don't have access to `.nvmrc` anymore.

.github/workflows/build-lint-test.yml
```diff
@@ -8,11 +8,17 @@ jobs:
...
+      - name: Install Yarn
+        run: corepack enable
+      - uses: actions/checkout@v4
+      - name: Use Node.js and install dependencies
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
```
mcmire, line 21: It looks like `corepack` may be already installed on `ubuntu-latest`? I'm reviewing solutions others have posted in the `actions/setup-node` ticket, specifically https://github.com/actions/setup-node/issues/480#issuecomment-1915448139, and it seems like we could do this without having to install Node twice:

``` suggestion
      - uses: actions/checkout@v4
      - name: Install Yarn via Corepack
        run: corepack enable 
      - name: Use Node.js and install dependencies
        uses: actions/setup-node@v4
        with:
          node-version: '.nvmrc'
```

.github/workflows/build-lint-test.yml
```diff
@@ -23,11 +29,17 @@ jobs:
     needs:
       - prepare
     steps:
+      - name: Use Node.js
```
mcmire, line 32: Similar as above.

.github/workflows/build-lint-test.yml
mcmire: I was also traversing the chain of various attempts at adding Corepack support to `setup-node`, and apparently according to [this comment](https://github.com/actions/setup-node/pull/482#pullrequestreview-1017248195) Yarn is already installed on GitHub runners somehow, so we may wish to remove it so that the shim that Corepack installs doesn't conflict:

``` suggestion
      - uses: actions/checkout@v4
      - name: Uninstall existing Yarn
        run: npm -g uninstall yarn
      - name: Install Yarn via Corepack
        run: corepack enable 
      - name: Use Node.js and install dependencies
        uses: actions/setup-node@v4
        with:
          node-version: '.nvmrc'
```

Then again, this doesn't make any sense since it would imply that `npm` is always preinstalled on GitHub runners. Just a suggestion though in case it works.

mcmire:
Thanks! I'll go ahead and merge this into my branch.

PR #245, opened 2024-04-26
chore: unpin yarn@^4.1.1

legobeat:
Should be reasonable to allow later yarn 4.x versions.

Related:
- #243 
- #244


mcmire:
Hmm, lame.

PR #246, opened 2024-06-11
Fix other workflows to install Corepack as prereq

mcmire:
A recent commit upgraded the version of Yarn to v4 and removed the Yarn binary from the repo, thereby requiring that Corepack be installed in order to install dependencies. The `build-lint-test` was updated to install Corepack, but not the documentation-related workflows, so they are failing.

This commit fixes those workflows to install Corepack. It also cleans up some work that was done in previous commits:

- In the `build-lint-test` workflow we ensure that `prepare` is run once per Node version we are testing and that `build` and `lint` use the latest Node version we are testing.
- In steps where we are installing Node just to gain access to the `corepack` executable, we use `.nvmrc` (the version of Node we use for development) to know which version of Node to install rather than using the latest LTS (`lts/*`). For jobs that do not need to concern themselves with being run in multiple Node versions, this ensures that consistent Node versions are used in this step vs. the step that is used to simply restore the Yarn cache.
- The checkout step always goes first, this way `.nvmrc` can be read.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -7,41 +7,47 @@ jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.x, 20.x]
     steps:
-      - name: Use Node.js
+      - uses: actions/checkout@v4
+      - name: Install Corepack via Node
         uses: actions/setup-node@v4
         with:
-          node-version: 'lts/*'
+          node-version-file: '.nvmrc'
       - name: Install Yarn
         run: corepack enable
-      - uses: actions/checkout@v4
-      - name: Use Node.js and install dependencies
+      - name: Install Node.js ${{ matrix.node-version }} and restore Yarn cache
         uses: actions/setup-node@v4
         with:
-          node-version: 'lts/*'
+          node-version: ${{ matrix.node-version }}
           cache: 'yarn'
-      - name: Install Yarn dependencies
+      - name: Install dependencies via Yarn
         run: yarn --immutable
 
   build:
     name: Build
+    needs: prepare
     runs-on: ubuntu-latest
-    needs:
-      - prepare
+    strategy:
+      matrix:
+        node-version: [20.x]
     steps:
-      - name: Use Node.js
+      - uses: actions/checkout@v4
+      - name: Install Corepack via Node
         uses: actions/setup-node@v4
         with:
-          node-version: 'lts/*'
+          node-version-file: '.nvmrc'
       - name: Install Yarn
         run: corepack enable
-      - uses: actions/checkout@v4
-      - name: Use Node.js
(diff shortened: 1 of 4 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -118,31 +127,32 @@ jobs:
...
-    needs:
-      - prepare
     strategy:
       matrix:
         node-version: [18.x, 20.x]
     steps:
-      - name: Use Node.js
+      - uses: actions/checkout@v4
+      - name: Install Corepack via Node
         uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
mcmire: I didn't find evidence in the [source](https://github.com/actions/setup-node) that it installs dependencies — I think you still have to do that yourself. 

[It _does_ run `yarn` to figure out how to cache Yarn files](https://github.com/actions/setup-node/blob/eff380dfbcf941bf8832e4acb788cebe13dfd758/src/cache-utils.ts#L44), and Yarn v1 is preinstalled on the GitHub runners, and [Yarn v1 throws if `packageManager` refers to a non-1.x release](https://github.com/yarnpkg/yarn/pull/9009). So you're right about that. But... `setup-node` should only try to run `yarn` if you pass the `cache` option to `setup-node`. I left that off the first `setup-node` call as you'd done and I re-run `setup-node` with caching after installing Corepack. So I think that should be okay, but do you still see a case in which this could fail?

As for your final point:

> Could potentially cause issues with any dependencies that rely on the specific package manager version in their installation step

Does Yarn consult `packageManager` for dependencies? I wasn't aware that it did that. I thought it only consulted it for the root package.

.github/workflows/build-lint-test.yml
```diff
@@ -7,41 +7,47 @@ jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.x, 20.x]
```
legobeat, line 12:
```suggestion
        node-version: [18.x, 20.x, 22.x]
```

.github/workflows/build-lint-test.yml
```diff
@@ -87,26 +96,26 @@ jobs:
 
   test:
     name: Test
+    needs: prepare
     runs-on: ubuntu-latest
-    needs:
-      - prepare
     strategy:
       matrix:
         node-version: [18.x, 20.x]
```
legobeat, line 103:
```suggestion
        node-version: [18.x, 20.x, 22.x]
```

.github/workflows/build-lint-test.yml
```diff
@@ -118,31 +127,32 @@ jobs:
 
   compatibility-test:
     name: Compatibility test
+    needs: prepare
     runs-on: ubuntu-latest
-    needs:
-      - prepare
     strategy:
       matrix:
         node-version: [18.x, 20.x]
```
legobeat, line 134:
```suggestion
        node-version: [18.x, 20.x, 22.x]
```

PR #247, opened 2024-06-13
Replace tsup with ts-bridge

mcmire:
Thanks to `tsup`, we publish CommonJS- and ESM-compatible versions of our libraries. However, there are two problems with our builds:

1. We produce two versions of JavaScript files, but we only produce one version of TypeScript type declaration files (which happens to be CommonJS-compatible). This violates best practices communicated both by from the TypeScript team and also the "Are the Types Wrong" tool, which both dictate that there be a 1-to-1 mapping between source and declaration files. You can read more about this problem here:
<https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md>
2. `tsup` identifies and extracts common code to "chunk" files. These make tree shaking possible, but make debugging difficult for our client teams.

This commit replaces `tsup` with `ts-bridge`, which solves both of these problems and more. It also adds `attw` to the `test` script so that we ensure that we don't have incompatibilities with builds in the future.

## References

Fixes https://github.com/MetaMask/metamask-module-template/issues/248.

## Examples

We've already done this in `utils`: https://github.com/MetaMask/utils/pull/182

## Testing

Run `yarn build` on `main`, and make a note of the files generated.

Now run `yarn build` on this branch. You should see a simplified file structure. No chunk files should exist, and `types/` should not exist. There should also be `.d.cts` files to match `.cjs` files and `.d.mts` files to match `.mjs` files.


rickycodes (approved):
This all makes sense to me

mcmire:
Notes on Socket report:

- `mz` is OK: it wraps `child_process` and `dns`. It was released 7 years ago, so it's old, but otherwise seems fine
- `esbuild` is OK: it seems to have code in it that downloads a version of ESBuild suitable for the current platform, which makes sense
- `tree-kill` is OK: it shells out to do its job
- `source-map` is OK: `loganfsmyth` is a [core maintainer of Babel](https://babeljs.io/team)

mcmire:
Not sure if this will work for me but just in case:

@SocketSecurity ignore npm/mz@2.7.0
@SocketSecurity ignore npm/esbuild@0.18.20
@SocketSecurity ignore npm/tree-kill@1.2.2
@SocketSecurity ignore npm/source-map@0.8.0-beta.0

mcmire:
@SocketSecurity ignore npm/mz@2.7.0
@SocketSecurity ignore npm/esbuild@0.18.20
@SocketSecurity ignore npm/tree-kill@1.2.2
@SocketSecurity ignore npm/source-map@0.8.0-beta.0

Mrtenz (approved):
Nice!

PR #253, opened 2024-09-27
chore: Use Node 22

rekmarks:
[As of October 29](https://github.com/nodejs/Release/tree/87462aac66fe3ce1ee1e3f812af83ccef9873782), Node.js 22 is the active `lts` version, and we should start to configure our repositories to use it as the latest supported version. This also updates our minimum supported Node versions for 18 and 20.

### DIFF
--- a/yarn.config.cjs
+++ b/yarn.config.cjs
@@ -179,7 +179,7 @@ async function expectPullRequestTemplate(workspace, workspaceName) {
 
   const pullRequestTemplate = await getWorkspaceFile(
     workspace,
-    '.github/PULL_REQUEST_TEMPLATE.md',
+    '.github/pull_request_template.md',
   );
 
   if (!pullRequestTemplate) {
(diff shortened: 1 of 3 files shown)


yarn.config.cjs
```diff
@@ -179,7 +179,7 @@ async function expectPullRequestTemplate(workspace, workspaceName) {
 
   const pullRequestTemplate = await getWorkspaceFile(
     workspace,
-    '.github/PULL_REQUEST_TEMPLATE.md',
+    '.github/pull_request_template.md',
```
rekmarks, line 182: Notably, whatever `fs` calls is backing this is newly case-sensitive in Node 22, so this fails if it doesn't actually match the filename.

legobeat:
Why not wait until it's actually lts? It's still `active` meaning breakages do happen (we had a couple under 22.x already) to a higher extent compared to `current` (20).

rekmarks:
Sure, we can wait for a month.

rekmarks:
Node.js 22.x is LTS [as of today](https://github.com/nodejs/Release/tree/87462aac66fe3ce1ee1e3f812af83ccef9873782).

MajorLift (approved):
LGTM!

mcmire (approved):
Looks good!

PR #254, opened 2024-11-04
Bump ESLint and related dependencies

Mrtenz:
ESLint 8 and the legacy config format are EOL. In this PR I've updated ESLint to v9, and bumped all related dependencies (ESLint configs, Prettier, `auto-changelog`), and rewritten the config to use the flat config format.


.prettierrc.mjs
```diff
@@ -1,8 +1,11 @@
 // All of these are defaults except singleQuote, but we specify them
 // for explicitness
-module.exports = {
+const config = {
```
mikesposito, line 3: nit: Do you know if the above comment still applies?

> // All of these are defaults except singleQuote, but we specify them
// for explicitness

eslint.config.mjs
```diff
@@ -0,0 +1,47 @@
+import base, { createConfig } from '@metamask/eslint-config';
+import jest from '@metamask/eslint-config-jest';
+import nodejs from '@metamask/eslint-config-nodejs';
+import typescript from '@metamask/eslint-config-typescript';
+
+const config = createConfig([
+  {
+    ignores: ['!.eslintrc.js', '!.prettierrc.js', 'dist/', 'docs/', '.yarn/'],
```
mikesposito, line 8: As we removed `.eslintrc.js` and renamed `.prettierrc.js`, should this list also be changed?

.prettierrc.mjs
Mrtenz: The values specified are still the defaults, with the exception of the `plugins` item (required for Prettier 3). We don't specify all options, but I'm not sure if that's necessary.

eslint.config.mjs
Mrtenz: Good catch! It seems like these are no longer ignored by default in ESLint 9, so I will remove them.

eslint.config.mjs
mcmire, line 1: I know that the `createConfig` function was already released and published. However, I can't help but wonder whether it will, in the long run, make our ESLint configurations unnecessarily difficult to understand and debug. ESLint already provides [recommendations for combining configuration objects (and exporting configs so they can be properly combined)](https://eslint.org/docs/latest/use/configure/combine-configs), and these are what plugins like `eslint-plugin-import-x`, `eslint-plugin-jsdoc`, and `eslint-plugin-prettier` follow in switching to the flat config format.

Upon switching to ESLint 9, I would expect the config file to look like the following:

``` javascript
import baseConfigs from '@metamask/eslint-config';
import typescriptConfigs from '@metamask/eslint-config-typescript';
import nodejsConfigs from '@metamask/eslint-config-nodejs';
import jestConfigs from '@metamask/eslint-config-jest';

export default [
  ...baseConfigs,
  {
    ignores: ['dist/', 'docs/', '.yarn/'],
  },
  {
    languageOptions: {
      sourceType: 'module'
    },
    settings: {
      'import-x/extensions': ['.js', '.mjs'],
    },
  },
  ...typescriptConfigs.map((config) => ({
    files: ['**/*.ts'],
    languageOptions: {
      parserOptions: {
        tsconfigRootDir: import.meta.dirname,
        project: ['./tsconfig.json'],
      },
    },
  }}),
  ...nodeJsConfigs.map((config) => ({
    ...config,
    files: ['**/*.js', '**/*.cjs'],
    languageOptions: {
      sourceType: 'script',
    },
  })),
  ...nodeJsConfigs.map((config) => ({
    ...config,
    files: ['**/*.test.ts', '**/*.test.js'],
    languageOptions: {
      sourceType: 'script',
    },
  })),
  ...jestConfigs.map((config) => ({
    ...config,
    files: ['**/*.test.ts', '**/*.test.js'],
  })),
];
```

While using `map` is a bit more verbose, I think it's very clear that we are creating objects to add to the flat array that we are ultimately exporting. I am not sure that is so clear with the `createConfig` utility.

What are your thoughts on this? I believe there are also further improvements we can make to the ESLint packages so that we can streamline this file, but I will leave that for a future comment.

Mrtenz: I considered this, and after talking about it with @rekmarks, we decided to create a simple util function that handles it for you. I don't think it changes so much in understandability of the config, and because it's less verbose, I feel like it's easier to read.

The function is based on [`typescript-eslint`'s config function](https://typescript-eslint.io/packages/typescript-eslint#flat-config-extends), so it's not like we're the only ones doing this.

> Upon switching to ESLint 9, I would expect the config file to look like the following:
>
> [...]

This gets very verbose if you actually want to override some rules or other settings. For example:

```ts
...typescript.map((config) => {
  rules: {
    ...config.rules,
    'some-rule': 'off',
  },

  languageOptions: {
    ...config.languageOptions,
    parserOptions: {
      ...config.languageOptions.parserOptions,
      ecmaVersion: 2023,
    }
  }
});
```

Of course we could use some deep merge function, but that's essentially what `createConfig` does, but easier.

> I can't help but wonder whether it will, in the long run, make our ESLint configurations unnecessarily difficult to understand and debug.

I don't think this will be any more of a problem with `createConfig` compared to `map` and spreading the config. Either way it's quite easy to log the object to see what's going on.

rekmarks: I find ESLint's recommendations, i.e. the `map()` syntax, completely unreadable, and much prefer the utility function.

mcmire (approved):

FrederikBolding (approved):

PR #257, opened 2025-01-15
Use artifacts instead of caching during release publishing

NEllusion:
## Summary

This pull request switches from using GitHub's `actions/cache` to GitHub artifacts as a mechanism to cache during builds. 

**Below is a screenshot showcasing passing builds in a private fork of this repo**

Screenshot

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "@metamask/eslint-config-jest": "^14.0.0",
     "@metamask/eslint-config-nodejs": "^14.0.0",
     "@metamask/eslint-config-typescript": "^14.0.0",
-    "@ts-bridge/cli": "^0.6.0",
+    "@ts-bridge/cli": "^0.6.3",
     "@types/jest": "^28.1.6",
     "@types/node": "^18.18",
     "@yarnpkg/types": "^4.0.0-rc.52",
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -33,14 +33,17 @@ jobs:
...
+          name: publish-release-artifacts-${{ github.sha }}
+          retention-days: 4
+          include-hidden-files: true
           path: |
             ./dist
-            ./node_modules/.yarn-state.yml
-          key: ${{ github.sha }}
-      - run: yarn --immutable
-      - run: yarn build
+            ./node_modules/
```
Gudahtt, line 46: We intentionally avoided running `yarn install` or including dependencies in the cache here in order to ensure third-party dependencies were not present during the step where the npm token is used, limiting risk of exfiltration.

Can we continue excluding it here? I am not following where it is needed

Gudahtt: It was indeed not intended. Looks like it was added here: https://github.com/MetaMask/metamask-module-template/pull/246
@mcmire I'm guessing that it was accidental to include third-party deps in the publish step, where they were previously excluded (I don't see this mentioned in the discussion)

mcmire: I think it was accidentally introduced, yes, sorry.

mcmire (approved):
LGTM.

Gudahtt (approved):
LGTM!

mcmire:
@NicholasEllul Are we good to merge this?

PR #258, opened 2025-01-15
Remove caching from publish docs workflow

NEllusion:
## Summary

Follow up to #257 - as we no longer cache our dependencies, we can remove code related to loading from cache. While this PR may look like we are removing the setup for node, we call the `setup-node` action a few lines before in this job.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -9,7 +9,6 @@ on:
         required: true
       PUBLISH_DOCS_TOKEN:
         required: true
-
 jobs:
   publish-release:
     permissions:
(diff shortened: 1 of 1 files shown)


mcmire:
Aren't we still caching dependencies in `build-lint-test.yml`?

Gudahtt:
@mcmire Yes, that's correct. I believe the intent was to still use caching in non-publishing workflows.

Gudahtt (approved):
LGTM!

mcmire (approved):

mcmire:
@NicholasEllul Are we good to merge this?

PR #260, opened 2025-02-26
Migrate to Vitest

Mrtenz:
This replaces Jest with Vitest. Vitest has better support for ESM (which we use in some repositories), works with Prettier 3, and generally "just works," without too much additional configuration.

## Examples

We are using Vitest in a few repos, including:

- [`ocap-kernel`](https://github.com/MetaMask/ocap-kernel)
- `create-release-pr`:  MetaMask/create-release-branch#167


src/index.test-d.ts
Mrtenz, line 1: This file demonstrates that Vitest has support for type tests out of the box. I figured it may be helpful to make developers aware of this.

src/index.test.ts
```diff
@@ -1,6 +1,8 @@
+import { describe, it, expect } from 'vitest';
```
Mrtenz, line 1: Vitest doesn't inject globals like `describe`, `it`, `expect` by default. We could enable a setting for this, but I think importing is preferred over using globals.

tsconfig.test.json
Mrtenz, line 1: Since Vite uses `esbuild` under the hood, tests aren't type checked by default. Using a separate `tsconfig` seemed to be the most reliable way to get type checking and tests working.

For example, if a test in a `test-d.ts` file fails, Vitest outputs the following error:

```
 FAIL  src/index.test-d.ts > greeter > returns a string
TypeCheckError: Type 'number' does not satisfy the constraint '"Expected number, Actual string"'.
 ❯ src/index.test-d.ts:7:49
      5| describe('greeter', () => {
      6|   it('returns a string', () => {
      7|     expectTypeOf(greeter('Huey')).toEqualTypeOf<number>();
       |                                                 ^
      8|   });
      9| });
```

And type errors in regular test files are detected as well, as unhandled source errors:

```
TypeCheckError: Type 'string' is not assignable to type 'number'.
 ❯ src/index.test.ts:7:11
      5| describe('greeter', () => {
      6|   it('greets', () => {
      7|     const name: number = 'Huey';
       |           ^
      8|     const result = greeter(name);
      9|     expect(result).toBe('Hello, Huey!');
```

tsconfig.test.json
```diff
@@ -0,0 +1,22 @@
...
+    "exactOptionalPropertyTypes": true,
+    "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2023"],
+    "module": "ES2022",
+    "moduleResolution": "Bundler",
+    "noEmit": true,
+    "noErrorTruncation": true,
+    "noUncheckedIndexedAccess": true,
+    "skipLibCheck": true,
+    "skipDefaultLibCheck": true,
```
Mrtenz, line 13: Vite's own declarations seem to rely on `"dom"` types. I don't think we should worry about Vite's declarations, and we can simply ignore checking those. These two options potentially speed up type checking slightly as well.

Mrtenz:
@SocketSecurity ignore npm/tinypool@1.0.2
@SocketSecurity ignore npm/tinyexec@0.3.2

These are expected to use `child_process` as they are used by Vitest.

@SocketSecurity ignore npm/esbuild@0.25.0

Network access is ok. Esbuild uses network access to download the right binary version depending on the platform.

@SocketSecurity ignore npm/istanbul-lib-report@3.0.1

New author is ok.

tsconfig.test.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "exactOptionalPropertyTypes": true,
+    "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2023"],
+    "module": "ES2022",
+    "moduleResolution": "Bundler",
```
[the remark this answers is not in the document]
Mrtenz: Vitest uses Vite to transpile TypeScript. `Bundler` is the most appropriate setting in this case and recommended by Vite.

tsconfig.test.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
```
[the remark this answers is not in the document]
Mrtenz: Makes sense! Updated.

tsconfig.test.json
```diff
@@ -0,0 +1,19 @@
...
+    "esModuleInterop": true,
+    "exactOptionalPropertyTypes": true,
+    "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2023"],
```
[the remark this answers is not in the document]
Mrtenz: Removed this now that it extends `tsconfig.json`, but to clarify, `lib` specifies the ECMAScript version to support, whereas `module` specifies  how to output the code (e.g., ESM/ES2022, or CommonJS).

tsconfig.test.json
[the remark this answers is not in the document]
mcmire: Okay, makes sense.

src/index.test-d.ts
mcmire: Yeah this is amazing, I had no clue!

mcmire:
@Mrtenz I'm seeing a peer dep warning when installing dependencies, is this important?

```
@metamask/metamask-module-template@workspace:. doesn't provide @typescript-eslint/utils (pdb740), requested by @vitest/eslint-plugin.
```

tsconfig.test.json
mcmire: Great!

Mrtenz:
> @Mrtenz I'm seeing a peer dep warning when installing dependencies, is this important?
> 
> ```
> @metamask/metamask-module-template@workspace:. doesn't provide @typescript-eslint/utils (pdb740), requested by @vitest/eslint-plugin.
> ```

It's installed regardless by other dependencies, but if it's a peer dependency, it makes sense to explicitly add it. We should update the installation instructions in `eslint-config` to mention this.

.depcheckrc.json
mcmire, line 1: This reminds me, we need to convert this file YAML so we can add comments to this file.

mcmire (approved):
Looks good!

.depcheckrc.json
Mrtenz: Sounds good, but out of scope for this PR

mcmire: Of course, I meant to say that. I created a ticket here: https://github.com/MetaMask/metamask-module-template/issues/265

FrederikBolding (approved):

PR #262, opened 2025-03-05
Implement `MetaMask/action-checkout-and-setup`

Mrtenz:
This implements `MetaMask/action-checkout-and-setup`, which replaces the checkout and Node.js setup steps in CI.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -11,7 +11,8 @@ jobs:
       matrix:
         node-version: [18.x, 20.x, 22.x]
     steps:
-      - uses: MetaMask/action-checkout-and-setup@v1
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
           node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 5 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -11,20 +11,12 @@ jobs:
...
-        uses: actions/setup-node@v4
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v1
         with:
+          is-high-risk-environment: false
           node-version: ${{ matrix.node-version }}
-          cache: 'yarn'
-      - name: Install dependencies via Yarn
-        run: yarn --immutable
+          cache-node-modules: ${{ matrix.node-version == '22.x' }}
```
FrederikBolding, line 19: Why do we only want to cache on a specific version?

Mrtenz: The cache key is the same, so it would show warnings that the cache could not be saved if we enable caching for all versions.

FrederikBolding: So do we not have caching for the other versions or we use the same cache?

Mrtenz: We use the same cache. AFAIK `node_modules` is the same for different Node.js versions.

FrederikBolding (approved):

mcmire (approved):
Looks good! Thanks for taking care of this.

PR #264, opened 2025-03-14
Remove unnecessary Yarn install from publish workflow

Mrtenz:
When migrating to `MetaMask/action-checkout-and-setup`, we accidentally forgot to remove this `yarn install`. It's redundant, since the workflow itself will handle it.


FrederikBolding (approved):

mcmire (approved):
Makes sense!

PR #267, opened 2025-06-06
Bump `eslint-plugin-import-x` in lockfile

Mrtenz:
This fixes an issue in the compatibility tests workflow in CI ([example here](https://github.com/MetaMask/metamask-module-template/actions/runs/15489104159/job/43610400347)) since the latest version of `eslint-plugin-import-x` uses `unrs-resolver`, which needs to be added to the `allow-scripts` configuration. I've bumped it in the lockfile here and added the `allow-scripts` configuration to resolve the CI error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
   "lavamoat": {
     "allowScripts": {
       "@lavamoat/preinstall-always-fail": false,
+      "eslint-plugin-import-x>unrs-resolver": false,
       "vite>esbuild": false
     }
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -260,6 +260,34 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@emnapi/core@npm:^1.4.3":
+  version: 1.4.3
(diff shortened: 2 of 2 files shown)


Gudahtt:
@SocketSecurity ignore npm/@emnapi/core@1.4.3
@SocketSecurity ignore npm/@tybys/wasm-util@0.9.0
@SocketSecurity ignore npm/@unrs/resolver-binding-wasm32-wasi@1.7.11
@SocketSecurity ignore npm/napi-postinstall@0.2.4
@SocketSecurity ignore npm/unrs-resolver@1.7.11

These packages have a really concerningly broad set of permissions 🤔. It does make sense given what they do, but it's really unfortunate that we need to use these for our ESLint plugin. We've crossed that bridge already so I won't suggest we reverse course here, and I understand this resolves some egregious performance issues in the original `eslint-plugin-import` package. But it's unfortunate. Ah well.

Gudahtt (approved):
LGTM!

cryptodev-2s (approved):
LGTM!

PR #268, opened 2025-08-08
chore: Add npm script for auto-fixing dependency lint issues

Gudahtt:
This matches what we use in `core`. Previously `lint` would actually run `yarn dedupe` and modify the lockfile, which is certainly not what I would expect from a lint command.

## Examples

https://github.com/MetaMask/core/blob/686a26ab125431f2f9cdb37ccbb28a96713e14f4/package.json#L27


Mrtenz (approved):
Makes sense.

PR #308, opened 2026-04-17
ci: Publish release to GitHub after NPM

Mrtenz:
This reorders the `publish-release` workflow to move the publish release to GitHub step after NPM publishing. NPM publish requires approval, and occasionally a release may be rejected and reverted, currently requiring manual removal of the Git tags and GitHub release, since those are always created before the approval.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the sequencing of the release pipeline (artifact build/publish dependencies), which could affect release automation if any job ordering assumptions were relied on.
> 
> **Overview**
> The `publish-release` GitHub Actions workflow is restructured so the build happens first, then NPM dry-run and NPM publish run from the built artifacts, and only **after NPM publish succeeds** does it create the GitHub release.
> 
> This avoids creating tags/releases before the `npm-publish` environment approval, and adds clearer job naming while keeping the docs publish steps (`gh-pages` and `latest`) dependent on the post-publish version/output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff690d0bce726ed932dca79a26bc9996ff85fa11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -36,9 +36,10 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
         with:
+          is-high-risk-environment: true
           ref: ${{ github.sha }}
       - name: Restore build artifacts
         uses: actions/download-artifact@v8
@@ -59,9 +60,10 @@ jobs:
     runs-on: ubuntu-latest
     environment: npm-publish
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
         with:
+          is-high-risk-environment: true
           ref: ${{ github.sha }}
       - name: Restore build artifacts
         uses: actions/download-artifact@v8
@@ -113,7 +115,7 @@ jobs:
       PUBLISH_DOCS_TOKEN: ${{ secrets.PUBLISH_DOCS_TOKEN }}
 
   publish-release:
-    name: Publish release to GitHub
+    name: Publish to GitHub
     needs: publish-npm
     permissions:
       contents: write
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -10,9 +10,8 @@ on:
       PUBLISH_DOCS_TOKEN:
         required: true
 jobs:
-  publish-release:
-    permissions:
-      contents: write
+  build:
+    name: Build
     runs-on: ubuntu-latest
     steps:
       - name: Checkout and setup environment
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release.yml
```diff
@@ -35,13 +32,13 @@ jobs:
...
 
   publish-npm-dry-run:
-    needs: publish-release
+    name: Publish to NPM (dry run)
+    needs: build
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout and setup environment
-        uses: MetaMask/action-checkout-and-setup@v3
+      - name: Checkout repository
```
Mrtenz: `MetaMask/action-npm-publish` doesn't need dependencies to be installed etc. from what I can tell, so a simple checkout is safer and quicker here.

.github/workflows/publish-release.yml
```diff
@@ -113,3 +111,19 @@ jobs:
...
       PUBLISH_DOCS_TOKEN: ${{ secrets.PUBLISH_DOCS_TOKEN }}
+
+  publish-release:
+    name: Publish release to GitHub
+    needs: publish-npm
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout and setup environment
```
Mrtenz, line 124: `MetaMask/action-publish-release` runs a Node.js script as composite action, so it potentially needs setup.

.github/workflows/publish-release.yml
Mrtenz: Tested this and it did not work without setting up Yarn.

.github/workflows/publish-release.yml
```diff
@@ -113,3 +113,19 @@ jobs:
...
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: true
+          ref: ${{ github.sha }}
+      - uses: MetaMask/action-publish-release@v3
```
mcmire, line 129: I see that this now goes after the doc publishing steps. Does it make sense for the doc publishing to go afterward instead?

This action also creates a Git tag in addition to creating the GitHub release. Do you think that will create a problem if the tag gets created after the package is published to NPM? I don't think so but I wanted to raise it in case it was.

Mrtenz: > I see that this now goes after the doc publishing steps. Does it make sense for the doc publishing to go afterward instead?

They run in parallel, `publish-release` only needs `publish-npm` (just like `get-release-version`, which is used for publishing the docs).

> This action also creates a Git tag in addition to creating the GitHub release. Do you think that will create a problem if the tag gets created after the package is published to NPM? I don't think so but I wanted to raise it in case it was.

Git and NPM tags are completely separate, so I don't think this should be a problem.

mcmire: Gotcha, thanks, makes sense then.

mcmire (approved):
LGTM.

GuillaumeRx (approved):

PR #313, opened 2026-06-03
ci: Enable NPM publishing with OIDC

Mrtenz:
This enables (staged) publishing to NPM with OIDC. Pending merge and release of this pull request:

- MetaMask/action-npm-publish#112.

---

> [!NOTE]
> **Medium Risk**
> Changes how packages are authenticated to npm at release time; misconfiguration could block publishes or briefly still depend on a long-lived NPM_TOKEN for bootstrap.
> 
> **Overview**
> Release CI is updated to publish to npm using **GitHub OIDC** via `MetaMask/action-npm-publish@v6`, with `id-token: write` on the reusable publish workflow caller and on the real publish job.
> 
> `NPM_TOKEN` is now **optional** at the workflow-call boundary; docs clarify it is only needed for the **first** publish to npm, after which OIDC can take over. The dry-run publish step no longer relies on omitting the token—it uses the same v6 action without passing `npm-token`.
> 
> Tooling bumps: **Yarn 4.16.0**, a pinned **`fflate@0.8.2`** resolution, and a minor `.yarnrc.yml` reorder of `nodeLinker`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7be3bab9b32e3642e2008bcadc739198ae6fe976. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


abretonc7s (dismissed):

package.json
```diff
@@ -81,6 +81,9 @@
     "vite": "^8.0.8",
     "vitest": "^4.1.4"
   },
+  "resolutions": {
+    "fflate": "0.8.2"
```
Mrtenz, line 85: I know we said we wouldn't do this, but it breaks the tests in the module template, making it impossible to merge anything. 😞

package.json
```diff
@@ -45,7 +45,6 @@
     "lint:eslint": "eslint .",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:constraints --fix && yarn lint:misc --write && yarn lint:dependencies:fix && yarn lint:changelog --fix",
     "lint:misc": "oxfmt --ignore-path .gitignore",
-    "prepack": "./scripts/prepack.sh",
```
Mrtenz, line 48: We already run `yarn build` in the `publish-release` workflow, so I don't think we need this?

Mrtenz: Restored this since it's somewhat out-of-scope for this pull request.

.github/workflows/publish-release.yml
```diff
@@ -50,8 +50,7 @@ jobs:
         with:
           name: publish-release-artifacts-${{ github.sha }}
       - name: Dry Run Publish
-        # omit npm-token token to perform dry run publish
-        uses: MetaMask/action-npm-publish@v5
+        uses: MetaMask/action-npm-publish@v6
```
FrederikBolding, line 53: How does the new version of this identify a dry-run if we don't always need the token?

Mrtenz: It checks if OIDC is available.

Mrtenz: https://github.com/MetaMask/action-npm-publish/blob/122fc82d6fdc62c647654d7c6726bf1f771c74bf/scripts/publish.sh#L42-L71

FrederikBolding: Hmm, okay. I wonder if we should have made that explicit when making a breaking change anyway.

FrederikBolding (approved):

PR #335, opened 2026-07-16
chore: apply `@lavamoat/harden` defaults

Mrtenz:
This implements the `@lavamoat/harden` defaults using `yarn dlx @lavamoat/harden defaults --level strict --package-manager yarn`. Changed include:

- The allow-scripts plugin was replaced with a more readable (unminified) version.
- A new runner plugin was added, which intercepts all package.json script calls and enables Node.js permissions.
   - The permissions are configured in the `lavamoat/scripts.*.json` files and the `scriptsConfig` field in package.json.

---

> [!NOTE]
> **Medium Risk**
> Changes how every workspace npm script executes (env, PATH, Node permissions) and install-time Yarn behavior; misconfigured profiles could break CI or local dev, but scope is repo tooling rather than runtime product code.
> 
> **Overview**
> Applies **@lavamoat/harden** strict defaults so `yarn run` scripts run under Node’s permission model, with Yarn install hardening alongside existing allow-scripts controls.
> 
> **Yarn** (`.yarnrc.yml`) turns on **hardened mode**, disables mirroring, and routes the global folder to `.yarn/global`. The bundled allow-scripts plugin under `.yarn/plugins` is removed in favor of local plugins: `lavamoat/plugin-allow-scripts.js` (still runs `yarn run allow-scripts` after install) and **`lavamoat/.runner-plugin.js`**, which hooks `wrapScriptExecution` for workspace scripts.
> 
> The **runner** loads per-script permission JSON from `package.json` **`scriptsConfig`** (default `lavamoat/scripts.strict.json`, with `lint`, `test:vitest`, etc. mapped to `scripts.lint.json`, `scripts.test.json`, `scripts.yarn.json`, and related profiles). It applies those options via `NODE_OPTIONS`, strips env vars matching keywords in `lavamoat/.env.ban.json`, and reorders `PATH` so `node_modules/.bin` is resolved last.
> 
> Supporting tweaks: split **`test`** / **`lint:dependencies`** into sub-scripts so each can get its own profile; **`test:package`** packs to `.yarn/package.tgz` for `attw`; add **`devEngines`** for Yarn 4.16+; ignore **`lavamoat/`** in ESLint/oxfmt; bump `@metamask/auto-changelog` and lockfile checksums.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 228265ecae1ba1d1b170268def9dbe4c71116963. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.depcheckrc.json
+++ b/.depcheckrc.json
@@ -10,6 +10,7 @@
     "@vitest/coverage-istanbul",
     "@vitest/eslint-plugin",
     "@yarnpkg/types",
+    "@yarnpkg/shell",
     "eslint-config-*",
     "eslint-import-resolver-typescript",
     "eslint-plugin-*",
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,6 +7,9 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
+# Enable hardened mode to validate lockfile content against remote registries.
+enableHardenedMode: true
+
 enableScripts: false
 
 enableTelemetry: false
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,10 +7,11 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
-enableMirror: false
-
 # Enable hardened mode to validate lockfile content against remote registries.
 enableHardenedMode: true
+
+# Disable Yarn's mirror feature to avoid writing to the global cache.
+enableMirror: false
 
 enableScripts: false
 
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -28,6 +28,9 @@ npmPreapprovedPackages:
   - '@metamask-previews/*'
   - '@lavamoat/*'
 
+# Protect the runtime of calls to "yarn run" scripts using a local plugin.
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
-    spec: 'https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js'
+    spec: https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js
+  - path: ./lavamoat/plugin-allow-scripts.js
+  - path: ./lavamoat/.runner-plugin.js
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -30,7 +30,5 @@ npmPreapprovedPackages:
 
 # Protect the runtime of calls to "yarn run" scripts using a local plugin.
 plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
-    spec: https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js
   - path: ./lavamoat/plugin-allow-scripts.js
   - path: ./lavamoat/.runner-plugin.js
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -17,6 +17,10 @@ enableScripts: false
 
 enableTelemetry: false
 
+# Despite `enableGlobalCache` and `enableMirror` being `false`, Yarn is still
+# writing to the global folder.
+globalFolder: .yarn/global
+
 logFilters:
   - code: YN0004
     level: discard
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,6 +7,8 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
+enableMirror: false
+
 # Enable hardened mode to validate lockfile content against remote registries.
 enableHardenedMode: true
 
--- a/lavamoat/.env.ban.json
+++ b/lavamoat/.env.ban.json
@@ -1,0 +1,1 @@
+["SESSION", "SSH", "KEY", "TOKEN", "SECRET", "PASSWORD", "AUTH"]
--- a/lavamoat/.runner-plugin.js
+++ b/lavamoat/.runner-plugin.js
@@ -17,6 +17,7 @@
 module.exports = {
   name: '@yarnpkg/plugin-runner',
   factory: function (/** @type {NodeJS.Require} */ require) {
+    const { tmpdir } = require('node:os')
     return {
       hooks: {
         /** @type {WrapScriptExecutionHook} */
@@ -25,25 +26,26 @@ module.exports = {
           project,
           locator,
           scriptName,
-          extra,
+          extra
         ) => {
-          const path = require('node:path');
-          const fs = require('node:fs');
-          const workspace = project.tryWorkspaceByLocator(locator);
+          const path = require('node:path')
+          const fs = require('node:fs')
+          const workspace = project.tryWorkspaceByLocator(locator)
 
           if (!workspace) {
-            // a script is being executed outside of a workspace context, so we can't apply any custom logic
+            // a script is being executed outside of a workspace context, so we
+            // can't apply any custom logic.
             // This is the case when a postinstal is running.
 
             // "Do nothing" - return the original executor immediately
             // without running any custom plugin logic or reading manifests.
             // TODO: implement wrapping these scripts with reasonable defaults
-            return executor;
+            return executor
           }
 
-          const pkgJson = workspace.manifest.raw;
+          const pkgJson = workspace.manifest.raw
           const binFolder =
-            extra.env.BERRY_BIN_FOLDER || `node_modules${path.sep}.bin`;
+            extra.env.BERRY_BIN_FOLDER || `node_modules${path.sep}.bin`
 
           const wrapper = makeRunScriptWrapper(
             {
@@ -51,31 +53,36 @@ module.exports = {
               scriptPayload: extra.script,
               projectRoot: extra.cwd,
               pathBinMatcher: (fragment) => {
-                return fragment.endsWith(binFolder);
+                return fragment.endsWith(binFolder)
               },
               customizePermissionsConfig: addMandatoryReads,
               readScriptsConfig: () => {
-                return pkgJson.scriptsConfig;
+                return pkgJson.scriptsConfig
               },
             },
             {
               readFileSync: fs.readFileSync,
               pathJoin: path.join,
               pathDelimiter: path.delimiter,
-            },
-          );
+              tmpdir,
+              realpathSync: fs.realpathSync,
+            }
+          )
 
-          const newEnv = wrapper.processEnv(extra.env);
+          // extra.env is a reference to the mutable object, but a different variable
+          // containing that reference is used within execute, so we must amend not
+          // replace it.
+          const newEnv = wrapper.processEnv(extra.env)
           for (const key of Object.keys(extra.env)) {
-            delete extra.env[key];
+            delete extra.env[key]
           }
-          Object.assign(extra.env, newEnv);
-          return executor;
+          Object.assign(extra.env, newEnv)
+          return executor
         },
       },
-    };
+    }
   },
-};
+}
 
 /**
  * @param {Record<string, boolean | string | string[]>} configOptions
@@ -83,30 +90,13 @@ module.exports = {
  */
 function addMandatoryReads(configOptions, _env) {
   if (!configOptions['--permission']) {
-    return;
+    return
   }
-  if (!Array.isArray(configOptions['--allow-fs-read'])) {
-    configOptions['--allow-fs-read'] = [];
-  }
-  if (!Array.isArray(configOptions['--allow-fs-write'])) {
-    configOptions['--allow-fs-write'] = [];
-  }
-
-  // figure out the /tmp dir for the current platform
-
-  const tmpdir =
-    process.platform === 'win32'
-      ? process.env.TEMP || '' // make typescript shut up
-      : '/tmp';
-  // yarn script execution makes heavy use of temporary dirs
-  configOptions['--allow-fs-read'].push(tmpdir);
-  configOptions['--allow-fs-write'].push(tmpdir);
+  configOptions['--allow-fs-tmp'] = true // yarn always uses tmp dirs.
 }
-/// <reference path="./makeRunScriptWrapper.global.d.ts" />
 
-/**
- * @typedef {Record<string, boolean | string | string[]>} ConfigOptions
- */
+;;
+/// <reference path="./makeRunScriptWrapper.global.d.ts" />
 
 /**
  * @param {MakeRunScriptWrapperOptions} param0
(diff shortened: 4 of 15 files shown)


.yarnrc.yml
```diff
@@ -28,6 +28,9 @@ npmPreapprovedPackages:
   - '@metamask-previews/*'
   - '@lavamoat/*'
 
+# Protect the runtime of calls to "yarn run" scripts using a local plugin.
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
-    spec: 'https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js'
+    spec: https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js
```
naugtur: you can now remove the old version of plugin-allow-scripts. The new one is the exact same functionality but readable.

package.json
```diff
@@ -51,7 +51,8 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.15.3",
-    "@lavamoat/allow-scripts": "^3.0.4",
+    "@lavamoat/allow-scripts": "^5.0.2",
+    "@lavamoat/harden": "^0.4.0",
```
naugtur: It's not necessary to have harden as a dependency. It's a standalone tool, you can installit via `npm i -g` and use on anything.

package.json
```diff
@@ -93,5 +101,8 @@
       "@lavamoat/preinstall-always-fail": false,
       "eslint-plugin-import-x>unrs-resolver": false
     }
+  },
+  "scriptsConfig": {
+    "#default": "lavamoat/scripts.loose.json"
```
naugtur: we could add some stronger policies for specific scripts here. 
I'll give it a go later.

lavamoat/plugin-allow-scripts.js
naugtur, line 1: Would you prefer for this file to also strt with a dot? I'd consider it an inconsistency on my side.

Mrtenz: I don't think `runner-plugin` needs to start with a dot, since it's already inside a separate `lavamoat` folder. But either way is fine with me.

lavamoat/scripts.lint.json
```diff
@@ -0,0 +1,14 @@
+{
+  "notes": "Denies all powerful IO capabilities that a linter script should not need. If a script needs more permissions, copy this file and edit.",
```
naugtur:
```suggestion
  "notes": "Denies all powerful IO capabilities that a linter script should not need. If a script needs more permissions, copy this file and edit. Do not allow more for all other linter scripts.",
```

lavamoat/scripts.strict.json
```diff
@@ -0,0 +1,14 @@
+{
+  "notes": "Denies all powerful IO capabilities that a linter script should not need. If a script needs more permissions, copy this file and edit.",
```
naugtur:
```suggestion
  "notes": "Denies all powerful IO capabilities. This provides a string default and if you need a script to have more permissions, choose another scripts.*.js file instead of editing this one.",
```

lavamoat/scripts.yarn.json
```diff
@@ -0,0 +1,14 @@
+{
+  "notes": "Denies all powerful IO capabilities that a linter script should not need. If a script needs more permissions, copy this file and edit.",
```
naugtur:
```suggestion
  "notes": "Use this one ONLY for scripts that are only making nested yarn processes, e.g. a composition of multiple other scripts via `yarn something && yarn somethingelse`",
```

package.json
```diff
@@ -40,20 +40,23 @@
     "lint": "yarn lint:eslint && yarn lint:constraints && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
     "lint:changelog": "auto-changelog validate --formatter oxfmt",
     "lint:constraints": "yarn constraints",
-    "lint:dependencies": "depcheck && yarn dedupe --check",
-    "lint:dependencies:fix": "depcheck && yarn dedupe",
+    "lint:dependencies": "yarn lint:dependencies:depcheck && yarn lint:dependencies:dedupe --check",
+    "lint:dependencies:depcheck": "depcheck",
+    "lint:dependencies:dedupe": "yarn dedupe",
+    "lint:dependencies:fix": "yarn lint:dependencies:depcheck && yarn lint:dependencies:dedupe",
```
naugtur, line 46: looking at this gives me a feature idea. 
```js
"scriptsConfig": {
    "lint": "lavamoat/scripts.yarn.json",
    "lint:*": "lavamoat/scripts.lint.json",
    "lint:constraints": "lavamoat/scripts.yarn.json",
```

`lint:*` would cover any beginsWith matches that don't have an exact match. 

That would save us a few lines while not introducing more complex syntax for one-to-many matching.

package.json
```diff
@@ -93,5 +103,18 @@
       "@lavamoat/preinstall-always-fail": false,
       "eslint-plugin-import-x>unrs-resolver": false
     }
+  },
+  "scriptsConfig": {
```
naugtur, line 108: the field is named `scriptsConfig` so that it lands after `scripts` when ordered alphabetically, so it's hard to miss the relationship ;)

naugtur (dismissed):

package.json
Mrtenz: That's `oxfmt` reformatting the `package.json` automatically 😅. It places unknown fields at the bottom.

naugtur: I didn't anticipate that one. Gotta hijack an existing field :P 

Also, made me think - since we're in control of the script string being passed around too, I could accept the config name with a separator at the beginning of the script. 

```js
"scripts": {
  "lint": "{lavamoat/scripts.lint.json} eslint"
```

I like the way it looks. It does make the default a little weird. And eliminates the ability to do prefixes for groups from the comment above.

Mrtenz: Hmm, I think I prefer `scriptsConfig` over that 🤔

.yarnrc.yml
```diff
@@ -7,10 +7,20 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
+# Enable hardened mode to validate lockfile content against remote registries.
+enableHardenedMode: true
+
+# Disable Yarn's mirror feature to avoid writing to the global cache.
+enableMirror: false
```
Mrtenz, line 14: Even though we disable the global cache above, Yarn still mirrors the local cache to the global cache (meaning it's still writing there), which is disallowed by the runner plugin in certain cases, e.g., when running `yarn dedupe` in a script.

.yarnrc.yml
```diff
@@ -7,10 +7,20 @@ compressionLevel: mixed
...
+
 enableScripts: false
 
 enableTelemetry: false
 
+# Despite `enableGlobalCache` and `enableMirror` being `false`, Yarn is still
+# writing to the global folder.
+globalFolder: .yarn/global
```
Mrtenz, line 22: Yarn always instantiates the PNPM linker class (I think?), which does some setup in the constructor including creating folders in the Yarn global folder. To get around this, the global folder is now stored locally.

package.json
```diff
@@ -40,20 +40,24 @@
     "lint": "yarn lint:eslint && yarn lint:constraints && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
     "lint:changelog": "auto-changelog validate --formatter oxfmt",
     "lint:constraints": "yarn constraints",
-    "lint:dependencies": "depcheck && yarn dedupe --check",
-    "lint:dependencies:fix": "depcheck && yarn dedupe",
+    "lint:dependencies": "yarn lint:dependencies:depcheck && yarn lint:dependencies:dedupe --check",
+    "lint:dependencies:depcheck": "depcheck",
+    "lint:dependencies:dedupe": "yarn dedupe",
+    "lint:dependencies:fix": "yarn lint:dependencies:depcheck && yarn lint:dependencies:dedupe",
```
Mrtenz, line 46: This split of scripts makes it possible to control permissions more granularly. All "orchestrator" scripts (scripts that only call other scripts like "lint", "lint:dependencies", "test") are set to use the "scripts.yarn.json" config, and all others use a more specific config.

package.json
```diff
@@ -40,20 +40,24 @@
...
     "lint:eslint": "eslint .",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:constraints --fix && yarn lint:misc --write && yarn lint:dependencies:fix && yarn lint:changelog --fix",
     "lint:misc": "oxfmt --ignore-path .gitignore",
     "prepack": "./scripts/prepack.sh",
-    "test": "vitest && attw --pack",
-    "test:watch": "vitest --watch"
+    "test": "yarn test:vitest && yarn test:package",
+    "test:vitest": "vitest",
+    "test:watch": "yarn test:vitest --watch",
+    "test:package": "yarn pack --out .yarn/package.tgz && attw .yarn/package.tgz"
```
Mrtenz, line 54: `attw --pack` internally calls `npm pack` which doesn't work with `devEngines` below. It blocks NPM from running completely.

lavamoat/.runner-plugin.js
Mrtenz, line 1: This plugin is added by `@lavamoat/harden`.

lavamoat/plugin-allow-scripts.js
Mrtenz, line 1: This plugin is added by `@lavamoat/harden`.

package.json
```diff
@@ -40,20 +40,24 @@
...
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.15.3",
     "@lavamoat/allow-scripts": "^5.1.0",
     "@lavamoat/preinstall-always-fail": "^2.0.0",
-    "@metamask/auto-changelog": "^6.1.0",
+    "@metamask/auto-changelog": "^6.2.1",
```
Mrtenz, line 60: Versions before `6.2.1` broke under the runner plugin because the binary didn't contain a shebang.

lavamoat/.runner-plugin.js
```diff
@@ -0,0 +1,370 @@
...
+
+/* global makeRunScriptWrapper */
+
+/**
+ * Yarn 4 plugin factory for wrapScriptExecution hook.
+ *
+ * @param {any} require - Yarn's require function
+ * @returns {{ hooks: { wrapScriptExecution: Function } }}
+ */
+module.exports = {
```
FrederikBolding, line 17: How do we keep these in sync now that they are just local copies?

Mrtenz: I don't think there's a good way to keep Yarn plugins in sync in general 😅. `@lavamoat/harden` has a `check` command which we can implement, but have to think about how to do it in a way that we don't break random PRs.

lavamoat/.env.ban.json
```diff
@@ -0,0 +1 @@
+["SESSION", "SSH", "KEY", "TOKEN", "SECRET", "PASSWORD", "AUTH"]
```
FrederikBolding, line 1: What is this

lavamoat/scripts.yarn.json
```diff
@@ -0,0 +1,15 @@
+{
+  "notes": "Use this one ONLY for scripts that are only creating nested Yarn processes, e.g., a composition of multiple other scripts via `yarn something && yarn something-else`.",
```
FrederikBolding, line 2: This seems a bit like a footgun tbh

lavamoat/.env.ban.json
Mrtenz: These are environment variables that are removed before running a script based on partial key match I believe.

lavamoat/scripts.yarn.json
Mrtenz: What do you mean? 🤔

lavamoat/scripts.yarn.json
```diff
@@ -0,0 +1,15 @@
...
+  "nodeOptions": {
+    "--disable-warning": "SecurityWarning",
+    "--permission": true,
+    "--allow-fs-read": ["/"],
+    "--allow-fs-write": ["./"],
+    "--allow-child-process": true,
```
FrederikBolding, line 8: Do child processes inherit the permissions of their parent?

lavamoat/scripts.yarn.json
FrederikBolding: Just that we need this to "glue" scripts together makes me worried that we accidentally give this permission to something we don't expect

lavamoat/scripts.yarn.json
Mrtenz: They do.

.depcheckrc.json
```diff
@@ -10,6 +10,7 @@
     "@vitest/coverage-istanbul",
     "@vitest/eslint-plugin",
     "@yarnpkg/types",
+    "@yarnpkg/shell",
```
FrederikBolding, line 13: I can't find a reference to this 🤔

Mrtenz: It's used in `./lavamoat/plugin-allow-scripts.js`.

lavamoat/.runner-plugin.js
FrederikBolding: How can I validate the contents of this one? 😄

.depcheckrc.json
FrederikBolding: Duh, I had that one collapsed 🫠

lavamoat/.runner-plugin.js
Mrtenz: You can run `yarn dlx @lavamoat/harden defaults --level strict --package-manager yarn` locally and diff. Currently there will be a one line difference to fix a bug we found. Zb is working on upstreaming it.

FrederikBolding (approved):

lavamoat/.runner-plugin.js
```diff
@@ -0,0 +1,370 @@
...
+            extra.env.BERRY_BIN_FOLDER || `node_modules${path.sep}.bin`
+
+          const wrapper = makeRunScriptWrapper(
+            {
+              scriptName,
+              scriptPayload: extra.script,
+              projectRoot: extra.cwd,
+              pathBinMatcher: (fragment) => {
+                return fragment.endsWith(binFolder)
+              },
```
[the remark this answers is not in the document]
naugtur: In my tests BERRY_BIN_FOLDER is the only path yarn adds to $PATH. So dependencies trying to run code via bin confusion could only be there.
Explain how `node_modules/.bin` would end up in PATH to make this a valid concern.

lavamoat/scripts.yarn.json
naugtur: In my (still short) experience, the package manager needs more permissions than many other things we do and it's unavoidable, so one thing we need to care about is having scripts that compose multiple `yarn something && yarn other` be cleanly separated from calling things that do actual things.

It could be possible to lint for that.

lavamoat/.env.ban.json
naugtur: more specifically - all env vars containing these words will be censored.

lavamoat/scripts.yarn.json
naugtur: the permissions are put in `NODE_OPTIONS` so as long as it's in the env, it gets passed along. 
programs can spawn processes with arbitrary env tho, so they can avoid passing that on. 

Normally in Node permissions model a subprocess doesn't inherit permissions

lavamoat/.runner-plugin.js
naugtur: the old way was not tracking whether the code is in sync anyway (correct me if I'm wrong) and was documenting trusting a github url, which is something we (and the overall community consensus) advise against these days.

I replaced the allow-scripts plugin built messy garbage from github url with a cleanly readable simple js file that everyone can audit at a glance. 
Runner is unfortunately a little more complex :shrug: 
We could make it work similarly to the allow-scripts plugin - by outsourcing all of the logic to a package and calling it from here. The problem is yarn docs say we should not require packages from dependencies in plugins (because plugins can obviously work before dependencies are installed, altho for this one it's unlikely) 
It doesn't change much imho. 

I'm hoping to stabilize the runner early and only add features via `--require` options in the json configs

PR #342, opened 2026-09-01
Replace PUBLISH_DOCS_TOKEN with Token Exchange Service token

mcmire:
Currently, to set up docs for a module template–derived repo, you must create a `metamaskbot` PAT for the repo and set it under `PUBLISH_DOCS_TOKEN`. However, `metamaskbot` PATs are deprecated in favor of the Token Exchange Service. This commit updates the docs publishing workflow (and dependent workflows) to make use of the Token Exchange Service.

---

> [!NOTE]
> **Medium Risk**
> Changes how production docs reach GitHub Pages; misconfigured Token Exchange policies could block publishes, though the workflow degrades with a clear warning rather than using a stale PAT.
> 
> **Overview**
> **Docs publishing no longer uses per-repo `PUBLISH_DOCS_TOKEN` secrets.** The reusable `publish-docs` workflow obtains a short-lived `contents: write` token via MetaMask’s `get-token` action and `vars.TOKEN_EXCHANGE_URL`, then passes it to `peaceiris/actions-gh-pages`.
> 
> Callers (`publish-main-docs`, `publish-rc-docs`, `publish-release`, and `main`) drop the `PUBLISH_DOCS_TOKEN` secret wiring and add `id-token: write` where they invoke docs publish. Checkout, build, and deploy run only when token exchange succeeds; if it fails (`continue-on-error`), the job emits a workflow warning about registering a Token Exchange policy instead of failing opaquely.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abf614e0239b72261acd8e511cd2e5c971e08b39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/publish-docs.yml
```diff
@@ -31,8 +35,8 @@ jobs:
       - name: Deploy to `${{ inputs.destination_dir }}` directory of `gh-pages` branch
         uses: peaceiris/actions-gh-pages@de7ea6f8efb354206b205ef54722213d99067935
         with:
-          # This `PUBLISH_DOCS_TOKEN` needs to be manually set per-repository.
-          # Look in the repository settings under "Environments", and set this token in the `github-pages` environment.
-          personal_token: ${{ secrets.PUBLISH_DOCS_TOKEN }}
+          # A policy must be registered for this repo and workflow
```
mcmire, line 43: I haven't added a policy for this repo, so with this change, the docs publishing workflow will no longer work. I'm okay with this considering this repo is just a template, but I can add it if we feel that seeing an X when viewing history for this repo would be confusing. Or maybe we can add a conditional around this workflow so that it doesn't run for template repos?

.github/workflows/publish-docs.yml
```diff
@@ -6,9 +6,6 @@ on:
       destination_dir:
         required: true
         type: string
-    secrets:
-      PUBLISH_DOCS_TOKEN:
-        required: true
```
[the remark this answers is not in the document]
mcmire: I knew I missed something. Fixed here: [ec2ba1e](https://github.com/MetaMask/metamask-module-template/pull/342/commits/ec2ba1ed89fcc9274af0e7f1459408d9e0951cf0)

.github/workflows/publish-docs.yml
Mrtenz: Should we make it exit with code 0 if it fails to get a token?

mcmire: If the `get-token` action fails, wouldn't it cause the workflow to fail too?

Mrtenz: Yeah, that's why I'm suggesting we add an exit 0 step if the get-token step failed 😅

mcmire: Ah sorry I misread your comment 😅

Yeah, that makes sense. I forgot that sometimes teams don't want to publish docs for their library, and if that's the case they shouldn't be penalized for it.

What do you think about something like this? [63173fb](https://github.com/MetaMask/metamask-module-template/pull/342/commits/63173fbafdf5632a367dd6cfa7f3488606f7f910)

Mrtenz: Looks good!

.github/workflows/publish-docs.yml
```diff
@@ -17,22 +14,38 @@ jobs:
...
-          personal_token: ${{ secrets.PUBLISH_DOCS_TOKEN }}
+          # A policy must be registered for this repo and workflow
+          # within the Token Exchange Service.
+          personal_token: ${{ steps.get-token.outputs.token }}
           publish_dir: ./docs
           destination_dir: ${{ inputs.destination_dir }}
+      - name: Show warning if the get-token step failed
+        if: ${{ steps.get-token.outcome == 'failed' }}
+        run: |
+          echo "::warning::The docs publishing workflow failed because the required token could not be obtained. If you want to publish docs for this repo, you'll need to add a policy to the Token Exchange Service."
```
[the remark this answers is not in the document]
mcmire: Oops, yes. Fixed in [abf614e](https://github.com/MetaMask/metamask-module-template/pull/342/commits/abf614e0239b72261acd8e511cd2e5c971e08b39).

Mrtenz (approved):

